PR #12, opened 2026-04-16
fix: make board cards keyboard accessible

matthewvryanjh:
## Summary
- make board cards focusable from the keyboard
- add Enter/Space activation for opening a board
- add a regression test covering keyboard activation

## Test Plan
- bun test src/extensions/Board/components/__tests__/BoardCard.test.tsx
- browser verification that the boards page now exposes an `Open board Bug Board` link with keyboard semantics

## Issue
- Real product bug found during production-ops validation: board cards on the Boards page were mouse-only and could not be opened by keyboard users.


rush1506:
@copilot Code is missing message internationalise using en.json and is using an unsupported test suite from the one that the repos is using.

rush1506:
@copilot we are not using that test suite, removed it. 

While this makes the card focusable and keyboard-activatable, which fixes the mouse-only access issue.

But Not ideal: role="link" should generally activate on Enter only. Space activation is button behavior.

Better options:
- Keep role="link" and trigger only on Enter.
- If you want Enter + Space, use role="button" (or better, a real button)

Best semantic fix if this navigates: render a real anchor/link element.

rush1506 (approved):

PR #14, opened 2026-04-22
fix(webhooks): allow owner update/delete for global webhooks

matthewvryanjh:
## Summary
- fix webhook `PATCH /api/v1/webhooks/:id` and `DELETE /api/v1/webhooks/:id` authorization for global webhooks
- remove stale workspace membership/admin checks from update/delete handlers
- add `canManageWebhook()` helper and wire both handlers through it
- add unit tests for owner vs non-owner authorization decision

## Root cause
After webhooks were made global (`workspace_id` nullable), update/delete still called `requireWorkspaceMembership(...)` using `webhook.workspace_id`, which rejects with `insufficient-role` when `workspace_id` is null.

## Validation
- `bun test tests/unit/webhooks/webhookPermissions.test.ts`
- `bun test tests/integration/webhooks/api-crud.test.ts tests/unit/webhooks/webhookPermissions.test.ts`
- local API smoke:
  - create webhook -> 201
  - patch webhook -> 200
  - delete webhook -> 200

## Notes
- independent review run: no security concerns, no logic errors

### DIFF
--- a/server/extensions/webhooks/api/delete.ts
+++ b/server/extensions/webhooks/api/delete.ts
@@ -1,12 +1,8 @@
-// DELETE /api/v1/webhooks/:id — remove a webhook; caller must be owner or workspace ADMIN+.
+// DELETE /api/v1/webhooks/:id — remove a webhook; caller must be the owner.
 // webhook_deliveries are cascade-deleted by DB foreign key constraint.
 import { db } from '../../../common/db';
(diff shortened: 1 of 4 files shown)


rush1506:
This is rejected mate, because the webhook is intended to create for the workspace to use in general, so any who has workspace access should be able to manage it. I think what you want is personal webhook with personal scope that works only for you, this would be an extended feature for webhook.

PR #33, opened 2026-09-07
security: pin markdown-it and linkify-it fixes

matthewvryanjh:
## Scope
Pins the vulnerable transitive Markdown parser chain without adding it as a direct runtime dependency.

- `markdown-it` `14.1.1` → `14.3.1`
- `linkify-it` `5.0.0` → `5.0.2`
- records the constraints in `package.json#overrides`, and regenerates both `bun.lock` and `package-lock.json`

## Why overrides
Both packages are transitive through `@tiptap/pm` → `prosemirror-markdown`; their declared ranges accept these fixes. Pinning them avoids an unrelated Tiptap upgrade while ensuring Bun and npm resolve the same patched versions.

## Validation
- `bun run build:client`
- Markdown parser/linkifier smoke test
- `git diff --check`

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -93,6 +93,10 @@
       },
     },
   },
+  "overrides": {
+    "linkify-it": "5.0.2",
+    "markdown-it": "14.3.1",
+  },
   "packages": {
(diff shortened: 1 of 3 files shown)


hermesrobinson-dev (approved):
Approved after independent review: scope is limited to compatible markdown-it/linkify-it security pins; both lockfiles resolve the patched versions and required verification is green.

PR #34, opened 2026-09-07
security: update sanitize-html to 2.17.7

matthewvryanjh:
## Scope
Updates the direct sanitiser dependency from `2.17.3` to `2.17.7`, with Bun and npm lockfiles regenerated.

## Runtime compatibility
`sanitize-html@2.17.7` declares Node `>=22.12.0`. Production uses Bun 1.3.5, so this PR validates the exact Docker dependency stage rather than assuming local compatibility.

## Validation
- `bun run build:client`
- sanitiser smoke test: strips a script tag while preserving allowed markup
- `docker build --target deps` using `oven/bun:1.3.5-slim`
- `git diff --check`

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -58,7 +58,7 @@
         "react-dom": "^18",
         "react-redux": "^9",
         "react-router-dom": "^6",
-        "sanitize-html": "^2.17.3",
+        "sanitize-html": "^2.17.7",
         "sharp": "^0.34.5",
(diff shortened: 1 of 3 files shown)


hermesrobinson-dev (approved):
Approved after independent review of the refreshed branch: scope is the direct sanitize-html security update, production Bun 1.3.5 dependency-stage validation passed, and required verification is green.

PR #35, opened 2026-09-07
ci: bootstrap scoped TypeScript linting

matthewvryanjh:
## Scope
Restores the missing `typescript-eslint` meta-package required by the existing flat ESLint config, then scopes strict type-aware linting to the production TypeScript project.

- strict type-aware lint: `src/**`, `server/**`, `vite.config.ts`, `tailwind.config.ts`
- ordinary TypeScript lint: tests, E2E, scripts, database utilities, CLI and root tooling outside the production tsconfig
- preserves strict production rules; does not disable them globally

## Result
- `bun run lint` now executes rather than failing at config load
- 215 TypeScript-project parsing failures removed
- remaining inventory: 5,188 real lint errors and 3 warnings, to be handled in small domain PRs

## Validation
- `bun run build:client`
- `git diff --check`

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -89,6 +89,7 @@
         "prettier": "^3.0.0",
         "tailwindcss": "^3",
         "typescript": "^5.6.0",
+        "typescript-eslint": "8.56.1",
         "vite": "^8",
       },
     },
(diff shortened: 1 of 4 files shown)


hermesrobinson-dev (approved):
Approved after independent review: restores the missing ESLint dependency and scopes strict type-aware rules to the production TypeScript project without globally weakening them. Required verification is green.

PR #36, opened 2026-09-07
fix(mcp): migrate create-board tool registration

matthewvryanjh:
## Scope
Migrates the `create_board` MCP tool from the deprecated `server.tool` API to `server.registerTool`.

The registered name, description, input schema, handler and API payload are unchanged. The tool test now asserts the new registration seam and fails explicitly if registration is absent.

## TDD evidence
- RED: test mock exposed only `registerTool`; existing `server.tool` implementation failed with `TypeError`
- GREEN: migrated implementation passes the same behaviour assertions

## Validation
- `bun test server/extensions/mcp/tools/createBoard.test.ts` — 2 passed
- focused ESLint — passed
- `git diff --check` — passed

Stacked on #35, the lint bootstrap PR.

### DIFF
--- a/server/extensions/mcp/tools/createBoard.test.ts
+++ b/server/extensions/mcp/tools/createBoard.test.ts
@@ -2,7 +2,7 @@ import { beforeEach, describe, expect, mock, test } from 'bun:test';
 
 const apiCallMock = mock();
 
-mock.module('../apiClient', () => ({
+void mock.module('../apiClient', () => ({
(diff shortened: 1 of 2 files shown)


hermesrobinson-dev (approved):
Approved after independent review: create_board now uses the supported MCP registerTool API; the regression test covered the old API failure and the focused test/lint checks pass. Required verification is green.

PR #37, opened 2026-09-07
fix(mcp): migrate create-list tool registration

matthewvryanjh:
## Scope
Migrates the `create_list` MCP tool from the deprecated `server.tool` API to `server.registerTool`.

The tool name, description, input schema, handler and API payload are preserved. The test now captures the supported registration config and fails explicitly when a handler is not registered.

## TDD evidence
- RED: test mock exposed only `registerTool`; the old `server.tool` implementation failed with `TypeError`
- GREEN: migrated implementation passes the same endpoint, payload and structured-error assertions

## Validation
- `bun test server/extensions/mcp/tools/createList.test.ts` — 2 passed
- focused ESLint — passed
- `git diff --check` — passed

### DIFF
--- a/server/extensions/mcp/tools/createList.test.ts
+++ b/server/extensions/mcp/tools/createList.test.ts
@@ -2,7 +2,7 @@ import { beforeEach, describe, expect, mock, test } from 'bun:test';
 
 const apiCallMock = mock();
 
-mock.module('../apiClient', () => ({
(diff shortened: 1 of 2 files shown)


hermesrobinson-dev (approved):
Approved after independent review: the create_list MCP registration migration preserves the public tool contract, has focused red-green coverage, and the required verification is green.

PR #38, opened 2026-09-07
fix(mcp): migrate get-card tool registration

matthewvryanjh:
## Scope
Migrates the `get_card` MCP tool from deprecated `server.tool` to `server.registerTool`.

Adds focused coverage for registration, GET endpoint/payload, success response and structured errors. The tool name, description, input schema and handler contract are preserved.

## TDD evidence
- RED: new `registerTool`-only fixture failed on the old `server.tool` call
- GREEN: migrated implementation passes both success and rejection cases

## Validation
- `bun test server/extensions/mcp/tools/getCard.test.ts` — 2 passed
- focused ESLint — passed
- `git diff --check` — passed

### DIFF
--- a/server/extensions/mcp/tools/getCard.test.ts
+++ b/server/extensions/mcp/tools/getCard.test.ts
@@ -1,0 +1,77 @@
+import { beforeEach, describe, expect, mock, test } from 'bun:test';
+
+const apiCallMock = mock();
+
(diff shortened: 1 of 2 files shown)


hermesrobinson-dev (approved):
Approved after independent review: get_card now uses the supported MCP registerTool API, with focused success and structured-error coverage. Required verification is green.

PR #39, opened 2026-09-07
fix(mcp): migrate create-card tool registration

matthewvryanjh:
## Scope
Migrates the `create_card` MCP tool from deprecated `server.tool` to `server.registerTool`.

Adds focused registration, endpoint/payload, success-response and structured-error coverage. The tool name, description, Zod input schema and API handler contract remain unchanged.

## TDD evidence
- RED: new `registerTool`-only fixture failed on the old `server.tool` call
- GREEN: migrated implementation passes both success and rejection cases

## Validation
- `bun test server/extensions/mcp/tools/createCard.test.ts` — 2 passed
- focused ESLint — passed
- `git diff --check` — passed

### DIFF
--- a/server/extensions/mcp/tools/createCard.test.ts
+++ b/server/extensions/mcp/tools/createCard.test.ts
@@ -1,0 +1,82 @@
+import { beforeEach, describe, expect, mock, test } from 'bun:test';
+
+const apiCallMock = mock();
+
+void mock.module('../apiClient', () => ({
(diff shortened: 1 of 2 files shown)


hermesrobinson-dev (approved):
Approved after independent review: create_card now uses the supported MCP registerTool API, preserving its schema and API handler contract with focused success and structured-error coverage.

PR #41, opened 2026-09-07
fix(mcp): migrate copy-state-transitions tool registration

matthewvryanjh:
## Scope
Migrates `copy_state_transitions` from deprecated `server.tool` to `server.registerTool`. Adds a focused registration-contract regression test; the existing input schema and handler remain unchanged.

## TDD evidence
- RED: registration-only fixture failed on `server.tool`
- GREEN: stable tool name, description and handler registration are preserved

## Validation
- `bun test server/extensions/mcp/tools/copyStateTransitions.test.ts` — passed
- focused ESLint — passed
- `git diff --check` — passed

### DIFF
--- a/server/extensions/mcp/tools/copyStateTransitions.test.ts
+++ b/server/extensions/mcp/tools/copyStateTransitions.test.ts
@@ -1,0 +1,35 @@
+import { beforeEach, describe, expect, test } from 'bun:test';
+
+let toolName = '';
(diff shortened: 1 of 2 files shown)


hermesrobinson-dev (approved):
Approved after independent review: copy_state_transitions uses the supported MCP registerTool API and preserves its registration contract.

PR #48, opened 2026-09-07
fix(mcp): migrate set-card-price tool registration

matthewvryanjh:
Migrates set_card_price from deprecated server.tool to server.registerTool. Adds a focused registration-contract regression test; the existing input schema and handler remain unchanged.


hermesrobinson-dev (approved):
Approved after independent review: set_card_price uses the supported MCP registration API and preserves its stable registration contract.

PR #50, opened 2026-09-07
fix(mcp): migrate write-comment tool registration

matthewvryanjh:
Migrates write_comment from deprecated server.tool to server.registerTool. Adds a focused stable registration-contract regression test; the existing input schema and handler remain unchanged.


hermesrobinson-dev (approved):
Approved after independent review: write_comment uses the supported MCP registration API and preserves its stable registration contract.

PR #51, opened 2026-09-07
fix(attachments): type attachment list rows

matthewvryanjh:
## Scope
Types the attachment-list database boundaries and extracts its pure response mapper. This removes the focused lint debt without changing the public attachment response contract.

## Validation
- focused mapper tests: 2 passed
- focused ESLint: passed
- client build: passed
- full Bun suite: 801 passed, 3 skipped, 118 failed, 93 errors; new mapper tests pass. Remaining failures are existing baseline debt.
- typecheck remains baseline-red outside this slice; no list.ts error is reported.
- git diff --check: passed

### DIFF
--- a/server/extensions/attachment/api/list.test.ts
+++ b/server/extensions/attachment/api/list.test.ts
@@ -1,0 +1,58 @@
+import { describe, expect, test } from 'bun:test';
+import { serializeAttachment } from './serializeAttachment';
+
+describe('serializeAttachment', () => {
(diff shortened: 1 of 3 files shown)


hermesrobinson-dev (approved):
Approved after independent review: attachment list database rows are typed at their query boundaries, while the extracted pure serializer preserves proxy-URL and referenced-card response behaviour under focused regression coverage.

PR #54, opened 2026-09-07
fix(attachments): type upload URL creation

matthewvryanjh:
## Summary
- type upload URL creation database boundaries
- preserve upload object key format through a pure helper and regression test

## Validation
- focused upload-key test
- focused ESLint
- existing MIME integration test remains baseline-red from 50 MB vs 100 MB assertion and invalid JWT fixture, outside this slice
- `bun run typecheck` (baseline-red; no upload-URL-slice diagnostics)
- `bun run build:client`
- `git diff --check`
- full `bun test` reporting: 804 pass, 3 skip, 118 fail, 93 errors (baseline suite debt; new test green)

### DIFF
--- a/server/extensions/attachment/__tests__/uploadKey.test.ts
+++ b/server/extensions/attachment/__tests__/uploadKey.test.ts
@@ -1,0 +1,10 @@
+import { describe, expect, test } from 'bun:test';
+import { buildUploadS3Key } from '../api/uploadKey';
+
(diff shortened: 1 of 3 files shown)


hermesrobinson-dev (approved):
Independent review: typed DB boundaries and actor guard preserve the presigned-upload validation and membership flow; upload-key contract is covered; required verify passed.

PR #56, opened 2026-09-07
fix(s3): split public and internal S3 endpoints

JaimeMarques:
## Problem

Browser uploads can use a public S3-compatible endpoint behind a CDN or reverse proxy. After a successful presigned PUT, ChimeDeck immediately checks the object using that same public endpoint. In the reproduced deployment, the immediate public `HeadObject` returned HTTP 403 while the same request to the direct object-store endpoint returned HTTP 200, so attachment confirmation incorrectly returned `upload-url-expired`.

## Approach

- Add optional `S3_INTERNAL_ENDPOINT`, falling back to `S3_ENDPOINT` for compatibility.
- Keep `s3Client` on the public endpoint for browser-facing presigned PUT/GET and multipart part URLs.
- Route direct server-side S3 operations through `s3ServerClient`: bucket setup, object HEAD/GET/PUT/DELETE, thumbnails, avatar/background uploads, and multipart create/complete/abort.
- Document the variable and the client split.
- Add an order-independent regression test that isolates the S3 singleton configuration in a child process.

## Verification

- `bun test server/extensions/attachment/common/config/s3.test.ts` — 1 passed, 0 failed. The isolated child covers four assertions.
- Sabotage check: routing `headObject` back through the public client makes the regression test fail; restoring the split makes it pass.
- `bunx eslint server/extensions/attachment/common/config/s3.test.ts` — passed.
- Prettier and `git diff --check` — passed.
- Client production build — passed.
- Full typecheck and integration suites remain red from pre-existing unrelated default-branch failures; baseline comparison found no new type errors and removed the prior S3 endpoint typing error.

## Compatibility and rollout

`S3_INTERNAL_ENDPOINT` is optional. Existing installations retain the old behavior when it is unset. Self-hosted deployments with a proxy/CDN in front of object storage should point it at their direct service endpoint, such as `http://object-store:9000`.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -25,6 +25,13 @@ SEED_FILE_URL=
 S3_REGION=us-east-1
 # Production AWS S3 endpoint (leave blank in local dev):
 S3_ENDPOINT=
+# Internal/direct endpoint for server-side S3 operations (optional).
+# Set when S3_ENDPOINT sits behind a public proxy: presigned URLs still target
+# S3_ENDPOINT, but direct server operations (Head/Delete/Get/Put, bucket
+# bootstrap, multipart control, thumbnails, avatars, board backgrounds) go
+# straight to this address — e.g. http://object-store:9000.
+# Falls back to S3_ENDPOINT when unset.
+S3_INTERNAL_ENDPOINT=
 # S3-specific credentials — set these to target LocalStack while keeping real AWS
 # credentials in AWS_ACCESS_KEY_ID / AWS_SECRET_ACCESS_KEY for SES.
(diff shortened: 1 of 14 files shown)


MrTheSoulz:
Closing to remove deployment-specific identifiers from the proposed change.

thanh-le-jh (approved):

hermesrobinson-dev (approved):
Independent review completed. Public presigning remains on S3_ENDPOINT; direct server S3 operations use the optional internal endpoint with a compatible fallback. PR merge ref passed the isolated S3 split regression test, focused ESLint and diff check; required verify is green.

PR #57, opened 2026-09-07
fix(attachments): encode Unicode content disposition

JaimeMarques:
## Problem

Attachment proxy responses placed the stored filename directly in the quoted `filename=` parameter. Runtime header validation rejects non-ASCII values, so otherwise valid files with Unicode names returned HTTP 500 when viewed or downloaded.

## Approach

- Add a pure RFC 6266/5987 `Content-Disposition` builder.
- Emit an ASCII-safe quoted `filename=` fallback.
- Emit `filename*=UTF-8''...` with the sanitized original filename percent-encoded for Unicode-capable clients.
- Strip control characters and path/quoted-string delimiters before constructing either value.
- Use the helper for both inline views and downloads.

## Verification

- Focused unit suite: 15 passed, 0 failed, 54 assertions.
- The test suite constructs real `Headers` and `Response` instances with Unicode filenames.
- RED verified before implementation because the helper was absent; the old raw-header behavior is covered by the real-header assertion.
- ESLint and Prettier passed for the new helper and test.
- Production client build and `git diff --check` passed.
- Deployment-identifier scan across the commit passed.

## Compatibility

ASCII filenames retain a readable quoted fallback. Unicode-aware browsers recover the sanitized original from `filename*`. Empty or unsafe-only names omit the optional header rather than constructing an invalid response.

### DIFF
--- a/server/extensions/attachment/common/contentDisposition.ts
+++ b/server/extensions/attachment/common/contentDisposition.ts
@@ -1,0 +1,38 @@
+// RFC 6266/5987: a quoted ASCII fallback plus the sanitized UTF-8 filename.
+// Strip controls and path/quoted-string delimiters before building either value.
+// eslint-disable-next-line no-control-regex
+const UNSAFE_FILENAME_CHARACTERS = /[\u0000-\u001F\u007F-\u009F"\\/]/g;
+const COMBINING_MARKS = /[\u0300-\u036F]/g;
+const RFC5987_EXTRA = /['()*]/g;
+
+export type ContentDispositionType = 'inline' | 'attachment';
+
+function toAsciiFallback(sanitized: string): string {
+  const folded = sanitized.normalize('NFD').replace(COMBINING_MARKS, '');
+  let fallback = '';
+  for (const ch of folded) {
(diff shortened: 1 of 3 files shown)


server/extensions/attachment/common/contentDisposition.ts
```diff
@@ -0,0 +1,38 @@
+// RFC 6266/5987: a quoted ASCII fallback plus the sanitized UTF-8 filename.
+// Strip controls and path/quoted-string delimiters before building either value.
+// eslint-disable-next-line no-control-regex
+const UNSAFE_FILENAME_CHARACTERS = /[\u0000-\u001F\u007F-\u009F"\\/]/g;
```
toan-nguyen-jh, line 4: LGTM. Optional UX nit: replace / and \ with _ instead of stripping them, so folder/report.pdf becomes folder_report.pdf rather than folderreport.pdf.

toan-nguyen-jh (approved):

hermesrobinson-dev (approved):
Independent review after rebase: the proxy retains s3ServerClient for direct internal reads, and adds a tested Unicode-safe RFC 6266/5987 Content-Disposition builder. Focused tests and lint passed; required verify is green.

PR #58, opened 2026-09-08
fix(attachments): type multipart upload start

matthewvryanjh:
## Summary
- type multipart-start database boundaries
- reuse the tested upload object-key contract

## Validation
- existing upload-key regression test
- focused ESLint
- multipart integration fixture remains baseline-red from invalid JWT test setup, outside this slice
- `bun run typecheck` (baseline-red; no multipart-start diagnostics)
- `bun run build:client`
- `git diff --check`
- full `bun test` reporting: 805 pass, 3 skip, 118 fail, 93 errors (baseline suite debt)


hermesrobinson-dev (approved):
Independent review: typed DB boundaries and actor guard preserve multipart upload validation, S3 initiation and pending-attachment creation; the established upload-key contract is reused; required verify passed.

PR #62, opened 2026-09-08
fix(attachments): clarify link preview timeout callback

matthewvryanjh:
## Summary
- make the link-preview timeout callback explicit to satisfy strict ESLint
- preserve timeout and abort behaviour

## Validation
- focused ESLint: passed
- client build: passed
- git diff --check: passed
- typecheck: baseline-red; no linkPreview.ts diagnostics
- full Bun suite: baseline unchanged (805 passed, 3 skipped, 118 failed, 93 errors)


hermesrobinson-dev (approved):
Independent review: one-file strict-lint clarification. The timeout still invokes AbortController.abort at the same delay; no API or behaviour change. Required verify passed and diff check is clean.

PR #70, opened 2026-09-08
fix(cards): type card label mutations

matthewvryanjh:
## Summary
- type card label context and ownership query boundaries
- preserve board-writable, membership and label-limit enforcement

## Validation
- bunx eslint server/extensions/card/api/labels.ts
- bun run build:client
- bun run typecheck (baseline-red; no labels.ts diagnostic)
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: scoped label mutation typing only. Board-writable, membership, ownership, idempotency and limit paths remain unchanged. verify passed.

PR #71, opened 2026-09-08
fix(cards): type card member mutations

matthewvryanjh:
## Summary
- type member assignment and removal query boundaries
- preserve membership, idempotency and activity behaviour

## Validation
- bunx eslint server/extensions/card/api/members.ts
- bun run build:client
- bun run typecheck (baseline-red; no members.ts diagnostic)
- git diff --check

### DIFF
--- a/server/extensions/card/api/members.ts
+++ b/server/extensions/card/api/members.ts
@@ -15,12 +15,34 @@ interface CardContext {
   workspaceId: string;
 }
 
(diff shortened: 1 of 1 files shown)


hermesrobinson-dev (approved):
Reviewed independently: scoped card-member typing only. Membership, idempotency and activity emission guards remain unchanged. verify passed.

PR #74, opened 2026-09-08
fix(cards): type card duplication

matthewvryanjh:
## Summary
- type writable-card, adjacent-card and duplicate query boundaries
- preserve positioning, copied fields and duplication event behaviour

## Validation
- bunx eslint server/extensions/card/api/duplicate.ts
- bun run build:client
- bun run typecheck (baseline-red; no duplicate.ts diagnostic)
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: scoped duplicate-card typing only. Writable-card access, positioning, copied-field and dispatch behaviour remain unchanged. verify passed.

PR #77, opened 2026-09-08
fix(cards): type card creation

matthewvryanjh:
## Summary
- type list, board and card query boundaries in card creation
- preserve board gating, append ordering, sanitisation, cover resolution, events and activity

## Validation
- bunx eslint server/extensions/card/api/create.ts
- bun run build:client
- bun run typecheck (baseline-red; no create.ts diagnostic)
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: one-file creation-route typed-boundary cleanup. Board access, append ordering, sanitisation, cover resolution, event and activity behaviour remain intact. The post-insert read now fails safely if absent. verify passed.

PR #81, opened 2026-09-08
fix(cards): type card archive mutation

matthewvryanjh:
## Summary
- type card, list, board and update-return boundaries in the archive route
- preserve direct archived-card handling, access checks, activity, realtime event and notification behaviour

## Validation
- bunx eslint server/extensions/card/api/archive.ts
- bun run build:client
- bun run typecheck (baseline-red; no archive.ts diagnostic)
- git diff --check

### DIFF
--- a/server/extensions/card/api/archive.ts
+++ b/server/extensions/card/api/archive.ts
@@ -11,12 +11,31 @@ import {
   type WorkspaceScopedRequest,
 } from '../../../middlewares/permissionManager';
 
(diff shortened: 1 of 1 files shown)


hermesrobinson-dev (approved):
Reviewed independently: pure typed DB/update-return boundary cleanup. Direct archived-card handling, access checks, toggle, activity, realtime event and notification sequence remain unchanged. verify passed.

PR #88, opened 2026-09-08
fix(boards): type starred board actor

matthewvryanjh:
## Summary
- narrow authenticated request after successful authentication in starred-board listing
- preserve board-star query and response behaviour

## Validation
- bunx eslint server/extensions/board/api/me-starred-boards.ts
- bun run build:client
- bun run typecheck (baseline-red; no me-starred-boards.ts diagnostic)
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: authenticated request narrowing only; query and response are unchanged. verify passed.

PR #91, opened 2026-09-08
fix(workspaces): type member removal

matthewvryanjh:
## Summary
- type target membership lookup in workspace member removal
- preserve ADMIN enforcement, member-not-found response, last-owner invariant and deletion behavior

## Validation
- bunx eslint server/extensions/workspace/api/members/remove.ts
- bun run build:client
- bun run typecheck (baseline-red; no members/remove.ts diagnostic)
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: typed target-membership lookup only. ADMIN checks, last-owner invariant, not-found response and deletion remain unchanged. verify passed.

PR #95, opened 2026-09-08
fix(boards): type board labels

matthewvryanjh:
## Summary
- narrow board contexts for label list/create routes
- align optional label color input type with its existing default
- preserve visibility, membership, guest and role authorization behavior

## Validation
- bunx eslint server/extensions/board/api/labels.ts
- bun run build:client
- bun run typecheck (baseline-red; no labels.ts diagnostic)
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: typed board contexts plus request-body type aligned to the existing colour fallback. Label visibility and creation authorisation remain unchanged. verify passed.

PR #96, opened 2026-09-08
fix(boards): type member removal

matthewvryanjh:
## Summary
- type board-member lookup and authenticated actor boundary in member removal
- preserve ADMIN enforcement, last-admin protection, deletion and event dispatch

## Validation
- bunx eslint server/extensions/board/api/members/remove.ts
- bun run build:client
- bun run typecheck (baseline-red; no members/remove.ts diagnostic)
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: typed board-member lookup and post-auth event actor only. ADMIN enforcement, last-admin protection, delete and event behavior remain unchanged. verify passed.

PR #97, opened 2026-09-08
fix(boards): type board follows

matthewvryanjh:
## Summary
- type authenticated-user and board lookup boundaries for idempotent board follow/unfollow
- preserve missing-board response, conflict-ignore follow and no-op unfollow behaviour

## Validation
- bunx eslint server/extensions/board/api/follow.ts
- bun run build:client
- bun run typecheck (baseline-red; no follow.ts diagnostic)
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: only authenticated-user and board lookup types were added to both follow paths. Missing-board and idempotent follow/unfollow behavior remains unchanged. verify passed.

PR #98, opened 2026-09-08
fix(boards): type board stars

matthewvryanjh:
## Summary
- type authenticated-user and board lookup boundaries for idempotent board star/unstar
- preserve missing-board response, conflict-ignore star and no-op unstar behaviour

## Validation
- bunx eslint server/extensions/board/api/star.ts
- bun run build:client
- bun run typecheck (baseline-red; no star.ts diagnostic)
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: only authenticated-user and board lookup typing changed in idempotent star/unstar paths. Missing-board handling and insert/delete behavior remain unchanged. verify passed.

PR #102, opened 2026-09-08
fix(boards): type member role update

matthewvryanjh:
## Summary
- type parsed body, member/count/readback rows and event actor in board member role updates
- preserve ADMIN authority, last-admin protection, update and event dispatch

## Validation
- bunx eslint server/extensions/board/api/members/update.ts
- bun run build:client
- bun run typecheck (baseline-red; no members/update.ts diagnostic)
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: request/database response typing only. ADMIN enforcement, last-admin guard, member role mutation and event dispatch remain unchanged. verify passed.

PR #104, opened 2026-09-08
fix(boards): type member suggestions

matthewvryanjh:
## Summary
- type board scope read boundary for member suggestions
- preserve private/workspace/public candidate scopes, guest inclusion, self exclusion and avatar response

## Validation
- bunx eslint server/extensions/board/api/members/suggestions.ts
- bun run build:client
- bun run typecheck (baseline-red; no members/suggestions.ts diagnostic)
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: only the board scope query is typed. Visibility-specific candidate scope, guest inclusion, self exclusion, filtering and avatar shaping remain unchanged. verify passed.

PR #105, opened 2026-09-08
fix(boards): type board archive

matthewvryanjh:
## Summary
- type board lookup and archive update return boundaries
- preserve membership, guest and ADMIN guards, state toggle and event dispatch

## Validation
- bunx eslint server/extensions/board/api/archive.ts
- bun run build:client
- bun run typecheck (baseline-red; no archive.ts diagnostic)
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: board lookup and update-return typing only. Access guards, ACTIVE/ARCHIVED toggle and board_archived event remain unchanged. verify passed.

PR #106, opened 2026-09-08
fix(boards): type background deletion

matthewvryanjh:
## Summary
- type writable board, background lookup and update return boundaries
- preserve access guards, non-fatal S3 cleanup, background clear and event dispatch

## Validation
- bunx eslint server/extensions/board/api/deleteBackground.ts
- bun run build:client
- bun run typecheck (baseline-red; no deleteBackground.ts diagnostic)
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: writable board and database boundaries are typed only. Access controls, non-fatal S3 cleanup, clear/update flow and board.background_changed dispatch remain unchanged. verify passed.

PR #107, opened 2026-09-08
fix(boards): type board deletion

matthewvryanjh:
## Summary
- type board lookup boundary for hard board deletion
- preserve confirmation requirements, access guards, deletion and workspace notification

## Validation
- bunx eslint server/extensions/board/api/delete.ts
- bun run build:client
- bun run typecheck (baseline-red; no delete.ts diagnostic)
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: this is a typed board read boundary only. Confirmation logic, membership/guest/ADMIN checks, deletion, workspace notification and response are unchanged. verify passed.

PR #108, opened 2026-09-08
fix(boards): type archived card retrieval

matthewvryanjh:
## Summary
- type middleware-established board scope and archived card query result
- preserve visibility/membership controls, label enrichment and response shape

## Validation
- bunx eslint server/extensions/board/api/archived-cards.ts
- bun run build:client
- bun run typecheck (baseline-red; no archived-cards.ts diagnostic)
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: board scope and joined-card result are typed only. Visibility/membership control flow, archived-card query, label enrichment and response remain unchanged. verify passed.

PR #109, opened 2026-09-08
fix(boards): type board retrieval

matthewvryanjh:
## Summary
- type resolved board and board-star lookup boundaries
- remove redundant null, async and promise aggregation operations
- preserve hydration, enrichment, access logging and response shape

## Validation
- bunx eslint server/extensions/board/api/get.ts
- bun run build:client
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: this preserves middleware access/logging flow, hydration, enrichments, activity inclusion, star query and response. The Promise.all removal is equivalent because the map body is synchronous. verify passed.

PR #111, opened 2026-09-08
fix(boards): type background proxy

matthewvryanjh:
## Summary
- type the board background query result
- preserve missing-background and external/S3 redirect paths

## Validation
- bunx eslint server/extensions/board/api/backgroundProxy.ts
- bun run build:client
- bun run typecheck (baseline-red; no backgroundProxy.ts diagnostic)
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: one typed background lookup only. Existing missing-background response, external redirect and short-lived S3 presigned redirect decisions are unchanged. verify passed.

PR #112, opened 2026-09-08
fix(boards): type board listing

matthewvryanjh:
## Summary
- type authenticated actor and each board-list query result
- remove unused role import and redundant response casts
- preserve role-specific visibility filters and board enrichment

## Validation
- bunx eslint server/extensions/board/api/list.ts
- bun run build:client
- bun run typecheck (baseline-red; no list.ts diagnostic)
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: authentication and role-specific visibility queries are unchanged. This only types authenticated actor/query results and removes an unused import and redundant casts. verify passed.

PR #113, opened 2026-09-08
fix(boards): type background upload

matthewvryanjh:
## Summary
- type writable-board, existing background and update readback boundaries
- preserve MIME/size checks, S3 upload/cleanup, event and response behaviour

## Validation
- bunx eslint server/extensions/board/api/uploadBackground.ts
- bun run build:client
- bun run typecheck (baseline-red; no uploadBackground.ts diagnostic)
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: resolved-board and DB readback typing only. Existing member authorisation, MIME/size validation, non-fatal prior-object cleanup, direct S3 upload, board update, event dispatch and response are unchanged. verify passed.

PR #116, opened 2026-09-08
fix(boards): type guest updates

matthewvryanjh:
## Summary
- type resolved board, request body, guest reads and event actor boundaries
- preserve guest validation, mutation, non-blocking event and response behaviour

## Validation
- bunx eslint server/extensions/board/api/guests/updateGuest.ts
- bun run build:client
- bun run typecheck (baseline-red; no updateGuest.ts diagnostic)
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: resolved board, body, guest-read and actor boundaries are typed only. ADMIN authorisation, validation, mutation, response and non-blocking event dispatch remain unchanged. verify passed.

PR #118, opened 2026-09-08
fix(boards): type board duplication

matthewvryanjh:
## Summary
- type the source-board lookup used by board duplication
- preserve authentication, member access, duplication and event behaviour

## Validation
- bunx eslint server/extensions/board/api/duplicate.ts
- bun run build:client
- bun run typecheck (baseline-red; no duplicate.ts diagnostic)
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: the source-board result is constrained to existing workspace_id/title reads after the not-found guard. Authentication, membership, guest/MEMBER checks, duplicate operation, failure mapping and event dispatch are unchanged.

PR #120, opened 2026-09-08
fix(workspaces): type workspace reads

matthewvryanjh:
## Summary
- type successful-authentication and workspace read boundaries
- preserve listing joins, membership enforcement and response mapping

## Validation
- bunx eslint server/extensions/workspace/api/get.ts
- bun run build:client
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: successful-authentication and workspace-read result boundaries are typed only. Caller-role listing, workspace membership enforcement, not-found response and response fields remain unchanged.

PR #122, opened 2026-09-08
fix(workspaces): type invite inspection

matthewvryanjh:
## Summary
- type the invite inspection readback
- preserve unauthenticated access, expiry/used flags and response mapping

## Validation
- bunx eslint server/extensions/workspace/api/invite/inspect.ts
- bun run build:client
- bun run typecheck (baseline-red; no inspect.ts diagnostic)
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: the invite readback is constrained to the fields already consumed. The unauthenticated route, not-found response, expired/used calculations and output mapping are unchanged.

PR #123, opened 2026-09-08
fix(boards): type guest access handlers

matthewvryanjh:
## Summary
- remove an unreferenced private legacy userId invite handler (zero repository call sites)
- type active guest revoke/list board and grant boundaries
- retain exported email invite and update handlers

## Validation
- bunx eslint server/extensions/board/api/guests/index.ts
- bun run build:client
- bun run typecheck (baseline-red; no guests/index.ts diagnostic)
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: handleInviteGuestById was private and has zero repository call sites; the exported email invite and update handlers remain. Active revoke/list authorization, membership cleanup and response behaviour are unchanged apart from typed read boundaries.

PR #124, opened 2026-09-08
fix(notifications): type bulk deletion

matthewvryanjh:
## Summary
- type the successful-authentication boundary for bulk notification deletion
- preserve caller-scoped deletion and 204 response

## Validation
- bunx eslint server/extensions/notifications/api/deleteAll.ts
- bun run build:client
- bun run typecheck (baseline-red; no deleteAll.ts diagnostic)
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: the successful-authentication user boundary is explicit only. The query remains scoped to the authenticated user and response remains 204.

PR #125, opened 2026-09-08
fix(notifications): type board preference reads

matthewvryanjh:
## Summary
- type resolved board and authenticated-user boundaries for board-type preference reads
- preserve visibility access and board-to-user-to-default preference resolution

## Validation
- bunx eslint server/extensions/notifications/api/boardTypePreferences/get.ts
- bun run build:client
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: resolved board and authenticated-user boundaries are typed only. Visibility enforcement plus board/user/default preference cascade and response shape are unchanged.

PR #126, opened 2026-09-08
fix(notifications): type board preference reset

matthewvryanjh:
## Summary
- type resolved board and authenticated-user boundaries for board-type preference reset
- preserve caller-and-board scoped deletion and fallback semantics

## Validation
- bunx eslint server/extensions/notifications/api/boardTypePreferences/reset.ts
- bun run build:client
- bun run typecheck (baseline-red; no reset.ts diagnostic)
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: resolved board and authenticated-user boundaries are typed only. Visibility enforcement, caller-and-board scoped deletion, fallback behaviour and response remain unchanged.

PR #128, opened 2026-09-08
fix(notifications): type bulk read marking

matthewvryanjh:
## Summary
- type the middleware-established authenticated-user boundary for bulk notification read marking
- preserve the caller-scoped unread update, count and response

## Validation
- bunx eslint server/extensions/notifications/api/markAllRead.ts
- bun run build:client
- bun run typecheck (baseline-red; no markAllRead.ts diagnostic)
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: only the middleware-established authenticated-user type boundary changes. The caller-scoped unread update, count and response semantics remain unchanged.

PR #129, opened 2026-09-08
fix(notifications): type single read marking

matthewvryanjh:
## Summary
- type authenticated-user and notification ownership lookup boundaries for single read marking
- preserve 404, caller ownership 403, read update and response semantics

## Validation
- bunx eslint server/extensions/notifications/api/markRead.ts
- bun run build:client
- bun run typecheck (baseline-red; no markRead.ts diagnostic)
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: only authenticated-user and owned-notification read boundaries are typed. The 404/403 outcomes, read update and response remain unchanged.

PR #130, opened 2026-09-08
fix(notifications): type single unread marking

matthewvryanjh:
## Summary
- type authenticated-user and notification ownership lookup boundaries for single unread marking
- preserve 404, caller ownership 403, unread update and response semantics

## Validation
- bunx eslint server/extensions/notifications/api/markUnread.ts
- bun run build:client
- bun run typecheck (baseline-red; no markUnread.ts diagnostic)
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: only authenticated-user and owned-notification type boundaries change. The 404/403 outcomes, unread update and response semantics remain unchanged.

PR #131, opened 2026-09-08
fix(notifications): type notification listing

matthewvryanjh:
## Summary
- type authenticated-user and list-result boundaries for notification listing
- inline the existing filters to remove an incompatible generic query helper without changing filter order
- preserve joins, pagination, cursor, reaction aggregation and response shape

## Validation
- bunx eslint server/extensions/notifications/api/list.ts
- bun run build:client
- bun run typecheck (baseline-red; no list.ts diagnostic)
- git diff --check

### DIFF
--- a/server/extensions/notifications/api/list.ts
+++ b/server/extensions/notifications/api/list.ts
@@ -103,34 +103,6 @@ function buildNotificationItem({
   };
 }
 
-function applyNotificationListFilters({
-  query,
-  unreadOnly,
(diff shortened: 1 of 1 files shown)


hermesrobinson-dev (approved):
Reviewed independently: typed request/result boundaries and filter inlining preserve filter order. Joins, pagination, cursor semantics, reaction aggregation and response shape are unchanged.

PR #133, opened 2026-09-08
fix(notifications): type global preference reads

matthewvryanjh:
## Summary
- type authenticated-user and global-setting read boundaries
- preserve user scoping, opt-out default and response shape

## Validation
- bunx eslint server/extensions/notifications/api/globalPreference/get.ts
- bun run build:client
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: only authenticated-user and setting-row type boundaries change. User scoping, opt-out default and response shape remain unchanged.

PR #134, opened 2026-09-08
fix(notifications): type global preference updates

matthewvryanjh:
## Summary
- type authenticated-user, parsed body, existence check and returning-row boundaries
- preserve boolean validation, update-versus-insert behaviour, returned fields and response shape

## Validation
- bunx eslint server/extensions/notifications/api/globalPreference/update.ts
- bun run build:client
- bun run typecheck (baseline-red; no update.ts diagnostic)
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: only authenticated-user, parsed-body, existence and returning-row type boundaries change. Boolean validation, update-versus-insert behaviour, returned fields and response shape remain unchanged.

PR #135, opened 2026-09-08
fix(notifications): type preference reads

matthewvryanjh:
## Summary
- type authenticated-user and preference-row collection boundaries
- preserve user scoping, all-type expansion, enabled-channel defaults and response shape

## Validation
- bunx eslint server/extensions/notifications/api/preferences/get.ts
- bun run build:client
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: only authenticated-user and preference-row collection type boundaries change. User scoping, all-type expansion, enabled-channel defaults and response shape remain unchanged.

PR #137, opened 2026-09-08
fix(notifications): type board preference reads

matthewvryanjh:
## Summary
- type resolved board/user, participant checks and preference-row boundaries
- preserve board visibility, participant-aware default, user scoping and response shape

## Validation
- bunx eslint server/extensions/notifications/api/boardPreference/get.ts
- bun run build:client
- git diff --check


hermesrobinson-dev (approved):
Reviewed independently: only resolved board/user, participant-check and preference-row type boundaries change. Visibility, participant-aware default, user scoping and response shape remain unchanged.

PR #138, opened 2026-09-08
fix(notifications): type board preference updates

matthewvryanjh:
## Summary
- type resolved board/user, parsed body, preference/participant checks and returning-row boundaries
- preserve validation, field merging, participant-aware new-row defaults, upsert behaviour and response shape

## Validation
- bunx eslint server/extensions/notifications/api/boardPreference/update.ts
- bun run build:client
- bun run typecheck (baseline-red; no update.ts diagnostic)
- git diff --check

### DIFF
--- a/server/extensions/notifications/api/boardPreference/update.ts
+++ b/server/extensions/notifications/api/boardPreference/update.ts
@@ -9,6 +9,16 @@ interface PatchBody {
   notifications_enabled?: unknown;
   only_related_to_me?: unknown;
 }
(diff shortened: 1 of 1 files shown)


hermesrobinson-dev (approved):
Reviewed independently: only resolved board/user, parsed-body, preference/participant checks and returning-row type boundaries change. Validation, field merging, participant-aware defaults, upsert behaviour and response shape remain unchanged.

PR #140, opened 2026-09-08
fix(automation): type automation listing

matthewvryanjh:
## Summary
- replace the unsafe authenticated-user assertion with an explicit 401 boundary
- preserve feature gating, creator-scoped board automation listing, trigger/action loading, ordering and response shape

## Validation
- bunx eslint server/extensions/automation/api/list.ts
- bun run build:client
- bun run typecheck (baseline-red; no automation/api/list.ts diagnostic)
- git diff --check


hermesrobinson-dev (approved):
Independent review complete. The one-file change replaces an unsafe assertion with an explicit 401 fallback after authentication, while retaining feature gating and creator-scoped automation listing. CI is green.

PR #142, opened 2026-09-08
fix(automation): type automation retrieval

matthewvryanjh:
## Summary
- type the authenticated-user, automation, trigger and action query boundaries
- preserve feature gating, creator-only lookup, 404 semantics, action ordering and formatted response output

## Validation
- bunx eslint server/extensions/automation/api/get.ts
- bun run build:client
- bun run typecheck (baseline-red; no automation/api/get.ts diagnostic)
- git diff --check


hermesrobinson-dev (approved):
Independent review complete. The one-file change narrows authenticated-user and automation query boundaries while preserving feature gating, creator-only lookup, 404 behaviour, action ordering and response formatting. CI is green.

PR #150, opened 2026-09-08
fix(state-transitions): simplify cache invalidation

matthewvryanjh:
## Summary
- Remove the impossible event-type branch from state-transition cache invalidation.
- The input type already guarantees `state_transition_updated`, so every prior valid call still invalidates the same board cache.

## Validation
- `bunx eslint server/extensions/stateTransitions/enforcement/rules.ts`
- `git diff --check`
- `bun run typecheck` (baseline exit 2; no changed-route diagnostic)


hermesrobinson-dev (approved):
Independent review complete. The event input is statically restricted to `state_transition_updated`; removing the unreachable guard preserves cache invalidation for every valid call. CI is green.

PR #151, opened 2026-09-08
fix(state-transitions): type error helper

matthewvryanjh:
## Summary
- Make the state-transition error helper’s open-string runtime contract explicit.
- Preserve the existing response shape and all accepted error names, including names used outside the literal alias.

## Validation
- `bunx eslint server/extensions/stateTransitions/common/errors.ts`
- `git diff --check`


hermesrobinson-dev (approved):
Independent review complete. The helper already accepts error names outside the literal alias; this change makes its existing open-string contract explicit while preserving the response structure and all runtime values. CI is green.

PR #153, opened 2026-09-08
fix(board): clarify visibility selection callback

matthewvryanjh:
## Summary
- Expand the board-visibility radio callback to avoid a conditional void expression.
- Preserve disabled-state and selected-visibility behaviour.

## Validation
- `bunx eslint src/extensions/Board/containers/BoardSettings/VisibilitySelector.tsx`
- `git diff --check`


hermesrobinson-dev (approved):
Independent review complete. The radio callback still invokes onChange only when enabled and passes the same option value; the change only avoids a conditional void expression. CI is green.

PR #154, opened 2026-09-08
fix(card): clarify label removal callback

matthewvryanjh:
## Summary
- Explicitly discard the label-detach promise in the chip’s void-only callback.
- Preserve label removal, picker attach/detach, and disabled rendering behaviour.

## Validation
- `bunx eslint src/extensions/Card/containers/CardModal/sections/LabelsSection.tsx`
- `git diff --check`


hermesrobinson-dev (approved):
Independent review complete. The chip still exposes removal only when enabled and invokes the same onDetach(label.id) action; this change only makes the void callback’s promise handling explicit. CI is green.

PR #155, opened 2026-09-08
fix(card): clarify member modal callbacks

matthewvryanjh:
## Summary
- Expand the Card modal member assignment and close callbacks to avoid conditional void expressions.
- Preserve member removal, modal visibility, assignment and disabled-state behaviour.

## Validation
- `bunx eslint src/extensions/Card/containers/CardModal/sections/MembersSection.tsx`
- `git diff --check`


hermesrobinson-dev (approved):
Independent review complete. The callbacks retain the exact same modal visibility transitions; member assignment/removal props and disabled behaviour are untouched. CI is green.

PR #156, opened 2026-09-08
fix(board): type background picker callbacks

matthewvryanjh:
## Summary
- Make background-picker error and DOM callback boundaries explicit.
- Preserve file selection, upload/removal request behaviour, state updates, error fallback, input reset and disabled controls.

## Validation
- `bunx eslint src/extensions/Board/containers/BoardSettings/BackgroundPicker.tsx`
- `git diff --check`


hermesrobinson-dev (approved):
Independent review complete. This preserves upload/remove requests, board state updates, fallback errors, input reset and disabled controls; it only makes error access and void-only DOM callback boundaries explicit. CI is green.

PR #157, opened 2026-09-08
fix(board): type board list API boundary

matthewvryanjh:
## Summary
- Type the injected API once and reuse it across BoardListPage thunks.
- Preserve every list/create/archive/delete/duplicate/star/unstar request, response mapping, reducer and optimistic-update path.

## Validation
- `bunx eslint src/extensions/Board/containers/BoardListPage/BoardListPage.duck.ts`
- `git diff --check`


hermesrobinson-dev (approved):
Independent review complete. The typed helper only formalises the injected API contract. All seven existing thunk requests, response mappings, reducer cases, stale-response protection and optimistic deletion behaviour remain unchanged. CI is green.

PR #167, opened 2026-09-08
fix(plugins): type board plugin list rows

matthewvryanjh:
## Scope
- Formalise the board and board-plugin join result boundaries in `list.ts`.
- Preserve active-only filtering, workspace membership, the response shape and the explicit exclusion of `api_key` from the plugin payload.

## Validation
- `bunx eslint server/extensions/plugins/api/board-plugins/list.ts`
- No dedicated endpoint test exists in this repository; no test was invented for a behaviour-preserving typing slice.
- `bun run build:client`
- `git diff --check`

### DIFF
--- a/server/extensions/plugins/api/board-plugins/list.ts
+++ b/server/extensions/plugins/api/board-plugins/list.ts
@@ -7,11 +7,36 @@ import {
   type WorkspaceScopedRequest,
 } from '../../../../middlewares/permissionManager';
 
+type BoardRow = { workspace_id: string };
+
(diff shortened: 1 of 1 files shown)


hermesrobinson-dev (approved):
Independent review complete. Typed board and join-result boundaries preserve authentication, workspace membership, active-only selection, client response transformation and the explicit exclusion of the plugin HMAC api_key. CI is green; no dedicated endpoint test exists and broader baseline failures are separately recorded.

PR #168, opened 2026-09-08
fix(automation): type move all cards action

matthewvryanjh:
## Scope
- Correct the local `ActionContext` import path and formalise the transaction, action config, list and card row boundaries in `moveAllCards.ts`.
- Preserve same-board guards, non-archived selection, position order, target-list append semantics and per-card transactional update order.

## Validation
- `bunx eslint server/extensions/automation/engine/actions/list/moveAllCards.ts`
- `bun test tests/integration/automation/actions.test.ts` — baseline blocked by empty `DATABASE_URL`; reproduced on untouched `origin/main`
- `bun run build:client`
- `git diff --check`

### DIFF
--- a/server/extensions/automation/engine/actions/list/moveAllCards.ts
+++ b/server/extensions/automation/engine/actions/list/moveAllCards.ts
@@ -3,12 +3,15 @@
 
 import { z } from 'zod';
 import { between, HIGH_SENTINEL } from '../../../../list/mods/fractional';
-import type { ActionHandler, ActionContext } from '../../../../common/types';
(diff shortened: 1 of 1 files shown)


hermesrobinson-dev (approved):
Independent review complete. The corrected action-context import and typed transaction/query boundaries preserve config validation, same-board guards, active-card filtering, ascending source order, append-after-target positioning and per-card transactional updates. CI is green. The focused automation suite remains baseline-blocked by the empty DATABASE_URL and was reproduced on untouched main.

PR #175, opened 2026-09-09
test(health-check): type row render fixtures

matthewvryanjh:
## Scope
- Move `HealthCheckRow` render smoke tests to the repository’s Bun test API.
- Align the fixture with the current required `HealthCheck.expectedStatus` field.
- Preserve all four rendered-output assertions: service/URL, probing spinner, response time and timeout label.

## Validation
- `bunx eslint src/extensions/HealthCheck/components/__tests__/HealthCheckRow.test.tsx`
- `bun test src/extensions/HealthCheck/components/__tests__/HealthCheckRow.test.tsx` — baseline-blocked: all four cases fail before assertions with `ReferenceError: document is not defined`. Reproduced on current `main` before this test-only change.
- `git diff --check`

### DIFF
--- a/src/extensions/HealthCheck/components/__tests__/HealthCheckRow.test.tsx
+++ b/src/extensions/HealthCheck/components/__tests__/HealthCheckRow.test.tsx
@@ -1,6 +1,6 @@
 // Render smoke tests for HealthCheckRow.
 import { render, screen } from '@testing-library/react';
-import { describe, it, expect, vi } from 'vitest';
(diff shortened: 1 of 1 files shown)


hermesrobinson-dev (approved):
Independent review complete. The test keeps the same four rendered-output scenarios and no-removal callback seam. Switching from unresolved Vitest imports to Bun test primitives and adding the required expectedStatus fixture field are type-boundary changes only. CI is green; the pre-existing Bun DOM-environment failure remains documented.

PR #179, opened 2026-09-09
fix(automation): type sort-by-name action

matthewvryanjh:
## Scope
- Correct the stale automation `ActionContext` type import and formalise sortable-card query results in `sortByName.ts`.
- Preserve non-archived list selection, case-insensitive title ordering and fractional-position update sequence.

## Validation
- `bunx eslint server/extensions/automation/engine/actions/list/sortByName.ts`
- `bun test tests/integration/automation/actions.test.ts` — baseline-blocked before tests: empty `DATABASE_URL` (`ERR_INVALID_URL`).
- `git diff --check`

### DIFF
--- a/server/extensions/automation/engine/actions/list/sortByName.ts
+++ b/server/extensions/automation/engine/actions/list/sortByName.ts
@@ -2,11 +2,16 @@
 
 import { z } from 'zod';
 import { between, HIGH_SENTINEL } from '../../../../list/mods/fractional';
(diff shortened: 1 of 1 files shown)


hermesrobinson-dev (approved):
Independent review complete. The corrected ActionContext import and explicit sortable-card query result preserve non-archived filtering, case-insensitive title/null ordering, and fractional-position update sequencing.

PR #180, opened 2026-09-09
fix(users): type current-user endpoint

matthewvryanjh:
## Scope
- Formalise authenticated-request and `users` query-result boundaries in `server/extensions/users/api/me.ts`.
- Preserve GET/PATCH response envelopes, update fields, 400/404 responses and avatar proxy resolution.
- Return 401 only if authentication unexpectedly completes without attaching `currentUser`.

## Validation
- `bunx eslint server/extensions/users/api/me.ts`
- No dedicated executable coverage for `handleGetMe` / `handlePatchMe` exists in this repository.
- `git diff --check`


hermesrobinson-dev (approved):
Independent review complete. The patch preserves deny-first authentication, existing GET/PATCH response envelopes, permitted update fields, not-found handling and avatar proxy resolution. The only new runtime branch is a defensive 401 for an otherwise invalid post-auth state.

PR #186, opened 2026-09-09
fix(auth): type board visibility query results

matthewvryanjh:
## Scope
- Formalise database/request boundaries in `server/middlewares/boardVisibility.ts` only.
- Preserve the board visibility access matrix, board-first 404 ordering, public bypass, authentication/membership flow, guest grants and derived context lookup paths.

## Expected validation
- `bunx eslint server/middlewares/boardVisibility.ts`
- `bun test tests/integration/board/boardVisibility.test.ts`
- `bun run typecheck`
- `bun test`
- `bun run build:client`
- `git diff --check`


hermesrobinson-dev (approved):
Independent review complete. Typed query boundaries preserve board-first lookup, visibility policy, membership and guest access flow, and all response paths.

PR #188, opened 2026-09-09
fix(auth): type verify email user result

matthewvryanjh:
## Scope
- Formalise the verified-user persistence boundary in `server/extensions/auth/api/verifyEmail.ts` only.
- Preserve token validation/expiry, verification update, refresh-token persistence, cookie attributes and response JSON exactly.

## Validation
- `bunx eslint server/extensions/auth/api/verifyEmail.ts`
- No dedicated verify-email handler test exists in the repository.
- `bun run typecheck`
- `bun test`
- `bun run build:client`
- `git diff --check`


hermesrobinson-dev (approved):
Independent review complete. The typed persisted-user boundary and Max-Age string conversion preserve token expiry validation, verification update, refresh persistence, cookie attributes, avatar resolution and response shape.

PR #191, opened 2026-09-09
test(state-transitions): type put mock

matthewvryanjh:
## Scope
- Remove lint-invalid mock/test-double patterns in `server/extensions/stateTransitions/api/__tests__/put.test.ts` only.
- Preserve feature-flag, graph validation, persistence, list-sync, GET roundtrip, error and broadcast contracts.

## Validation
- Focused ESLint and PUT contract suite
- `bun run typecheck`, `bun test`, `bun run build:client`, `git diff --check`


hermesrobinson-dev (approved):
Independent review complete. Query mock semantics, feature flag getter, board-not-found path, graph/list-sync contracts, persistence and publication assertions are preserved.

PR #192, opened 2026-09-09
test(health): type status dot assertions

matthewvryanjh:
## Scope
- Resolve lint typing in `src/extensions/HealthCheck/components/__tests__/HealthCheckStatusDot.test.tsx` only.
- Preserve all five render/accessibility assertions.

## Validation
- Focused ESLint
- Focused test is baseline-blocked before assertions: `ReferenceError: document is not defined` from Testing Library render.
- No executable DOM/E2E coverage is available locally for this slice; no UI server is started because implementation/UI behaviour is unchanged.
- `bun run typecheck`, `bun test`, `bun run build:client`, `git diff --check`


hermesrobinson-dev (approved):
Independent review complete. The five HealthCheckStatusDot render/accessibility assertions are unchanged; this only aligns the test API import and matcher typing.

PR #194, opened 2026-09-09
fix(comment): type inline image markdown hooks

matthewvryanjh:
## Scope

Types the untrusted Markdown token/node/helper boundaries in `src/extensions/Comment/extensions/InlineImage.ts` without changing parsing or rendering behaviour.

## Behaviour preserved

- Only string `href`, `text`, and `title` values are accepted.
- Missing or empty `href` returns `null` from Markdown parsing.
- Node creation retains `{ src, alt, title }`, with absent alt/title represented as `null`.
- Missing image source renders an empty Markdown string.
- Title remains optional in standard Markdown image output.

## Validation

- `bunx eslint src/extensions/Comment/extensions/InlineImage.ts` — passed, 0 findings
- `git diff --check` — passed
- `bun run typecheck` — existing exit 2; no `InlineImage.ts` diagnostic
- `bun test` — existing exit 1
- `bun run build:client` — passed

## Coverage limitation

No focused `InlineImage` test exists in the repository. This patch changes no UI implementation or runtime contract; it introduces typed guards around already-existing dynamic Markdown hook inputs.

### DIFF
--- a/src/extensions/Comment/extensions/InlineImage.ts
+++ b/src/extensions/Comment/extensions/InlineImage.ts
@@ -1,10 +1,45 @@
 // Minimal inline-image Tiptap extension for use in the comment editor.
 // Intentionally self-contained — does not depend on @tiptap/extension-image
 // whose built dist is not included in the package at this version.
-import { Node, mergeAttributes } from '@tiptap/core';
+import {
+  Node,
+  mergeAttributes,
+  type JSONContent,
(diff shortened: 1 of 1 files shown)


hermesrobinson-dev (approved):
Independent review complete. The guarded Markdown hook inputs preserve the original string-only extraction, empty-source handling, node attributes, and optional-title rendering contract.

PR #196, opened 2026-09-09
fix(health): type create handler boundaries

matthewvryanjh:
## Scope

Types the existing authentication and database-row boundaries in `server/extensions/healthCheck/api/create.ts` while preserving the handler contract.

## Behaviour preserved

- Authentication before board visibility.
- Invalid JSON, missing/invalid name and URL, and 120-character name limit responses.
- SSRF-safe URL validation; only `UrlValidationError` maps to `422`.
- `preset`/`custom` handling and preset-key requirement.
- Existing numeric `expectedStatus` coercion and 100–599 integer validation, including null handling.
- Case-insensitive duplicate URL rejection per board.
- UUID insert fields and exact `201` response shape.

## Validation

- `bunx eslint server/extensions/healthCheck/api/create.ts` — passed, 0 findings
- `git diff --check` — passed
- `bun run typecheck` — existing exit 2; no touched-file diagnostic
- `bun test` — existing exit 1
- `bun run build:client` — passed

## Coverage limitation

No dedicated create-health-check handler test exists in the repository. This is a type-boundary-only change; no endpoint behaviour was intentionally changed.

### DIFF
--- a/server/extensions/healthCheck/api/create.ts
+++ b/server/extensions/healthCheck/api/create.ts
@@ -9,11 +9,26 @@ import { validateUrl, UrlValidationError } from '../common/validateUrl';
 
 const MAX_NAME_LENGTH = 120;
 
+type AuthenticatedUserRequest = AuthenticatedRequest & { currentUser: { id: string } };
+
+type HealthCheckRow = {
+  id: string;
+  board_id: string;
+  name: string;
+  url: string;
+  type: string;
+  preset_key: string | null;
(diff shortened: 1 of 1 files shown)


hermesrobinson-dev (approved):
Independent review complete. Authentication/visibility order, external numeric-status coercion, validation errors, duplicate detection, insert payload and 201 response shape are retained; this only types auth and database boundaries.

PR #211, opened 2026-09-09
fix(workspace): type invite validation lookup

matthewvryanjh:
## Scope
Type invite validation’s authoritative DB lookup and remove the existing no-op cache read with its stale comment/imports.

## Behaviour preserved
Validation still reads the database, returning the same not-found, expired, already-used, or valid invite results. The removed cache value was never inspected or returned.

## Verification
- Focused ESLint: 0 findings
- `bun test tests/integration/trelloCompat/organizations.test.ts`: 6 passed
- `git diff --check`: pass

No UI code changed.


hermesrobinson-dev (approved):
Independent review: approved. The patch types the DB lookup and removes an unused cache read only; all observable invite-validation outcomes remain database-authoritative and unchanged. Focused organisation coverage and CI verify passed.

PR #212, opened 2026-09-09
fix(activity): mark webhook dispatch as intentional async

matthewvryanjh:
## Scope
Explicitly mark member-webhook delivery as intentional fire-and-forget work.

## Behaviour preserved
Webhook delivery remains non-blocking: activity/event callers do not await it, and webhook dispatch payload/selection are unchanged.

## Verification
- Focused ESLint: 0 findings
- `bun test server/extensions/activity/mods/__tests__/createActivityEvent.test.ts`: 4 passed
- `git diff --check`: pass

No UI code changed.


hermesrobinson-dev (approved):
Independent review: approved. `void` documents the existing non-blocking webhook dispatch without changing selection, payload, timing or caller response flow. Focused activity fan-out coverage and CI verify passed.

PR #214, opened 2026-09-09
fix(ids): type short-id uniqueness lookup

matthewvryanjh:
## Scope
Type the optional short-ID uniqueness lookup and make the generated alphabet character explicitly total.

## Behaviour preserved
Short IDs remain eight random base-62 characters by default; uniqueness still retries up to 20 times against the same tables and throws the same exhaustion error.

## Verification
- Focused ESLint: 0 findings
- `bun test tests/integration/trelloCompat/boards.test.ts tests/integration/trelloCompat/lists.test.ts`: 20 passed
- `git diff --check`: pass

No UI code changed.

### DIFF
--- a/server/common/ids/shortId.ts
+++ b/server/common/ids/shortId.ts
@@ -8,7 +8,9 @@ export function generateShortId(length = SHORT_ID_LENGTH): string {
   const bytes = randomBytes(length);
   let out = '';
   for (let i = 0; i < length; i += 1) {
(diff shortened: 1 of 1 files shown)


hermesrobinson-dev (approved):
Independent review: approved. This patch preserves base-62 ID generation and the existing 20-attempt uniqueness contract while typing the optional lookup. Focused Trello board/list coverage and CI verify passed.

PR #215, opened 2026-09-09
fix(crypto): type decrypt output as UTF-8 string

matthewvryanjh:
## Scope
Explicitly decode the streamed AES-GCM plaintext bytes as UTF-8 before concatenating the final string, and add focused cryptographic contract coverage.

## Behaviour preserved
Cipher format, key validation, AES-256-GCM configuration, encryption, authentication-tag verification, returned plaintext and tamper rejection are unchanged.

## Verification
- Focused ESLint: 0 findings
- `bun test tests/unit/server/common/crypto.test.ts`: 2 passed (UTF-8 round trip and ciphertext tamper rejection)
- `bun test`: existing exit 1 — 1,119 pass, 3 skip, 180 fail, 30 errors; two additional passing focused tests
- `git diff --check`: pass

No UI code changed.

### DIFF
--- a/server/common/crypto.ts
+++ b/server/common/crypto.ts
@@ -51,5 +51,5 @@ export function decryptSecret({ ciphertext, hexKey }: { ciphertext: string; hexK
   const encrypted = buf.subarray(IV_BYTE_LEN, buf.length - TAG_BYTE_LEN);
   const decipher = createDecipheriv(ALGORITHM, key, iv);
   decipher.setAuthTag(tag);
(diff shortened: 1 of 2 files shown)


hermesrobinson-dev (approved):
Independent review: approved. The streamed AES-GCM plaintext was already interpreted as UTF-8 by string concatenation; this makes that conversion explicit. Focused round-trip and tamper-rejection coverage proves the cryptographic contract; CI verify passed.

PR #216, opened 2026-09-09
fix(cards): narrow resolved cover guard

matthewvryanjh:
## Scope
Remove a redundant optional chain after the existing attachment lookup guard in card-cover resolution.

## Behaviour preserved
Missing, mismatched, non-ready and keyless attachments still resolve to no cover; valid ready attachments retain the same proxy URL, GIF handling and aspect-ratio logic.

## Verification
- Focused ESLint: 0 findings
- No executable direct test for `resolveCoverImageUrls` exists; no UI code changed
- `git diff --check`: pass


hermesrobinson-dev (approved):
Independent review: approved. JavaScript short-circuiting guarantees `attachment.status` is read only when the optional `card_id` check did not return, so missing/mismatched/non-ready covers keep the same null-cover outcome. CI verify passed.

PR #217, opened 2026-09-09
fix(ids): type entity ID lookup boundary

matthewvryanjh:
## Scope
Type the shared entity-ID lookup row and remove its assertion.

## Behaviour preserved
Board, card, list, comment and attachment resolution still checks canonical ID or short ID and returns either the canonical ID or `null`.

## Verification
- Focused ESLint: 0 findings
- `bun test tests/integration/trelloCompat/actions.test.ts tests/integration/trelloCompat/boards.test.ts tests/integration/trelloCompat/lists.test.ts`: 26 passed
- `git diff --check`: pass

No UI code changed.

### DIFF
--- a/server/common/ids/resolveEntityId.ts
+++ b/server/common/ids/resolveEntityId.ts
@@ -4,13 +4,13 @@ async function resolveEntityId(
   tableName: 'boards' | 'cards' | 'lists' | 'comments' | 'attachments',
   identifier: string,
 ): Promise<string | null> {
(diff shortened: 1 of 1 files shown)


hermesrobinson-dev (approved):
Independent review: approved. Typed query rows remove only an assertion; table coverage, canonical-or-short-ID lookup conditions, canonical return value and null fallback are unchanged. Focused Trello compatibility coverage and CI verify passed.

PR #218, opened 2026-09-09
fix(mentions): type board scope lookup

matthewvryanjh:
## Scope
Type the board workspace/visibility lookup used to scope @mention resolution.

## Behaviour preserved
Missing boards still return no users. Private boards remain restricted to board members/guests; workspace/public boards remain restricted to non-guest workspace members plus board guests; nickname and UUID-token matching are unchanged.

## Verification
- Focused ESLint: 0 findings
- No executable direct test for `resolveNicknames` exists; no UI code changed
- `git diff --check`: pass

### DIFF
--- a/server/common/mentions/resolve.ts
+++ b/server/common/mentions/resolve.ts
@@ -14,6 +14,11 @@ interface User {
   avatar_url: string | null;
 }
 
+interface BoardMentionScope {
+  workspace_id: string;
+  visibility: string;
+}
+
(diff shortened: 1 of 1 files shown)


hermesrobinson-dev (approved):
Independent review: approved. This types the existing board scope projection only. Private vs workspace/public recipient visibility, guest inclusion, UUID/nickname matching and missing-board fallback are unchanged. CI verify passed.

PR #244, opened 2026-09-09
Type users/profile GET+PATCH handler DB row reads

matthewvryanjh:
## Scope

Small, cohesive server-side ESLint slice: `server/extensions/users/api/profile/{get,update}.ts`
(GET/PATCH `/api/v1/users/me`, full-profile variant with nickname). Non-payment.

## Change

Replace `any`-typed, non-null-asserted `db('users')` reads with a migration-derived
`UserRow` type (columns from `db/migrations/0002_auth.ts`, `0014_email_verification.ts`,
`0015_user_profile.ts`), matching the existing typed pattern already used in
`server/extensions/users/api/me.ts`.

- `currentUser!.id` → explicit `if (!currentUser) return 401` guard (was previously a
  non-null assertion; the guard makes the failure mode explicit instead of throwing).
- `.first()` / `.returning('*')` are typed against `UserRow` (and `Pick<UserRow, 'id'>`
  for the existence-only nickname-uniqueness check).
- Dropped the now-redundant `email_verified ?? false` fallback: migration 0014 makes the
  column `NOT NULL DEFAULT false`, so the type is `boolean`, not `boolean | undefined`.

No route, auth, or response-shape/API-contract changes.

## Tests

Added `server/extensions/users/api/profile/__tests__/profileHandlers.test.ts`, which
spawns a subprocess fixture (`__tests__/fixtures/profileHandlers.ts`) that mocks
`common/db`, `auth/middlewares/authentication`, and `common/avatar/resolveAvatarUrl`
in isolation (per repo convention, to avoid `mock.module` leakage into the shared
suite) and exercises the **real** `handleGetProfile` / `handleUpdateProfile` handlers:

- auth short-circuit (401)
- missing-`currentUser` guard (401, exact body)
- user-not-found (404)
- full projection incl. nullable `nickname`/`avatar_url` and boolean `email_verified`
- nickname-format rejection without touching the DB
- nickname-uniqueness conflict (409) with exact `where`/`whereNot` query shape
- successful update returning full projection with avatar-proxy URL substitution

This is durable handler-level coverage of the changed module, not a mock-only smoke test.
Live DB/router integration and browser/E2E coverage are **not** exercised here (none
existed before this slice either); that gap is pre-existing and unchanged.

## Verification (fresh BASE captured before edits, `origin/main` @ `30bea5aa`)

| Gate | BASE | HEAD | Result |
|---|---|---|---|
| `bun test` (full suite) | 1228 pass / 3 skip / 180 fail / 30 errors | 1229 pass / 3 skip / 180 fail / 30 errors | +1 pass (new test); fail/error **named identities reconciled — no new, none lost** (verified via full-multiline diff, timestamps normalized) |
| `bun run typecheck` | 147 diagnostics | 147 diagnostics | Byte-identical diagnostic set (`diff` empty) |
| Focused `eslint` on touched files | 18 errors (get.ts) + 18 errors (update.ts) pre-slice | 0 | Clean |
| `bun run build:client` | — | exit 0 | Passes (pre-existing >500kB chunk-size warning only) |
| `git diff --check` | — | clean | No whitespace issues |

Logs (unique per run):
- `/tmp/chimedeck-base-typecheck.log`, `/tmp/chimedeck-head-typecheck-244.log`
- `/tmp/chimedeck-base-test.log`, `/tmp/chimedeck-head-test-244.log`
- `/tmp/base-fails-244.txt`, `/tmp/head-fails-244.txt` (+ normalized variants)
- `/tmp/chimedeck-slice-eslint-244.txt`
- `/tmp/chimedeck-head-buildclient-244.log`

## Emitted-JS comparison (Bun `--target bun --external '*'` transpile, base vs head)

Not byte-identical — two real, intentional deltas (reported honestly, not erased):

1. `get.ts`/`update.ts`: `const { currentUser } = req;` → explicit
   `if (!currentUser) return 401` block. Behavior-preserving in the reachable
   path (authenticate() only returns without error when currentUser is set),
   but is a genuine defensive-guard addition replacing a non-null assertion.
2. `email_verified ?? false` → `email_verified`: erased fallback, verified
   equivalent for all real rows given the `NOT NULL DEFAULT false` migration
   constraint (0014_email_verification.ts).

No other emitted-JS differences.

## Boundaries

- No repo-wide `eslint --fix`, no suppression comments, no config/dependency/
  deployment changes.
- No payments/billing/subscription code touched.
- Not merging this PR — opening for separate review per standing policy.

### DIFF
--- a/server/extensions/users/api/profile/__tests__/fixtures/profileHandlers.ts
+++ b/server/extensions/users/api/profile/__tests__/fixtures/profileHandlers.ts
@@ -1,0 +1,172 @@
+import { strict as assert } from 'node:assert';
+import { mock } from 'bun:test';
+
+const calls: unknown[] = [];
+const state: {
+  authError: Response | null;
+  currentUser: { id: string } | null;
+  row: Record<string, unknown> | undefined;
+  updated: Record<string, unknown>[];
+} = { authError: null, currentUser: { id: 'user-1' }, row: undefined, updated: [] };
+
+void mock.module('../../../../../../common/db', () => ({
+  db(table: string) {
+    calls.push(table);
+    return {
+      where(filter: Record<string, unknown>) {
+        calls.push(['where', filter]);
+        return this;
+      },
+      whereNot(filter: Record<string, unknown>) {
+        calls.push(['whereNot', filter]);
+        return this;
+      },
+      update(values: Record<string, unknown>) {
+        calls.push(['update', values]);
+        return this;
+      },
+      returning(columns: string) {
+        calls.push(['returning', columns]);
+        return Promise.resolve(state.updated);
+      },
+      first() {
+        calls.push('first');
+        return Promise.resolve(state.row);
+      },
+    };
+  },
+}));
+void mock.module('../../../../../auth/middlewares/authentication', () => ({
+  authenticate(req: Request) {
+    calls.push(req);
+    (req as { currentUser?: unknown }).currentUser = state.currentUser;
+    return Promise.resolve(state.authError);
+  },
+}));
+void mock.module('../../../../../../common/avatar/resolveAvatarUrl', () => ({
+  buildAvatarProxyUrl({ userId, avatarUrl }: { userId: string; avatarUrl: string | null }) {
+    calls.push(['buildAvatarProxyUrl', userId, avatarUrl]);
+    return avatarUrl ? `/api/v1/users/${userId}/avatar` : null;
+  },
+}));
+
+const { handleGetProfile } = await import('../../get');
+const { handleUpdateProfile } = await import('../../update');
+
+// --- GET: auth short-circuit ---
+const getReq = new Request('http://localhost/api/v1/users/me');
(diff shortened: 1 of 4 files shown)


hermesrobinson-dev (approved):
Independent parent review of cfc282af9bdbaf1a6478fcbe7bfdbbc410550a69: typed GET/PATCH profile handler DB rows via migration-derived UserRow. Replaced currentUser! non-null assertions with explicit 401 guards (behaviour-preserving on the only reachable path since authenticate() sets currentUser or short-circuits) and erased a redundant ?? false now that email_verified is NOT NULL per migration 0014. Reran focused lint (0 errors) and regression:1 pass covering auth short-circuit, missing-user guard, 404, full projection, nickname validation, 409 uniqueness and successful update. Typecheck diagnostics byte-identical base/head (147=147). Full test pass/fail/skip totals reconciled, no lost passing identities. No live DB/router coverage claimed — fake DB isolated in subprocess. Required verify passed.

PR #254, opened 2026-09-09
lint(search): type workspace search row projections

matthewvryanjh:
Types the board/card search row projections in queryWorkspaceSearch.ts (BoardSearchRow/CardSearchRow, derived from board/card/short-id migrations), fixing 13 no-unsafe-* / no-unnecessary-condition findings. applyBoardAccessFilter's qb param is now Knex.QueryBuilder instead of ReturnType<typeof db>, removing implicit-any callback params from the visibility where-clauses. Two disclosed no-op behavioural deltas confirmed via emitted-JS diff: dropped a redundant includeArchived ?? false (destructured default already guarantees boolean) and redundant Number(rank) conversions (ts_rank_cd/real is already a JS number from node-postgres). Added a durable subprocess-isolated handler test (none existed before) covering role-based board access filtering (OWNER/ADMIN/MEMBER/VIEWER/GUEST branches), rank-sorted board+card merge, background URL proxying, and rank-field exclusion from the public response shape.

Verification:
- Focused ESLint: 0 errors/warnings on all 3 touched files
- bun run typecheck: byte-identical vs main
- bun test: 1253->1254 pass (my +1 test only), 3 skip/180 fail/30 errors unchanged
- bun run build:client: passed
- git diff --check: passed

Coverage limits: fake-DB-mock subprocess test, no live Postgres full-text search coverage (pre-existing gap). No payments/Stripe/billing/monetisation involved.

### DIFF
--- a/server/extensions/search/mods/__tests__/fixtures/queryWorkspaceSearch.ts
+++ b/server/extensions/search/mods/__tests__/fixtures/queryWorkspaceSearch.ts
@@ -1,0 +1,252 @@
+import { strict as assert } from 'node:assert';
+import { mock } from 'bun:test';
+
+// Real queryWorkspaceSearch under test. The shared db module is mocked here in a
+// subprocess so this fixture cannot leak fake Knex chains into (or be replaced by)
+// adjacent test files that mock '../../../common/db' differently.
+const calls: string[] = [];
+
+interface FakeRow {
+  [key: string]: unknown;
+}
+
+// Minimal permissive chain recorder: every method returns `this` (or a nested fake
+// for callback-based .where(fn)) so the real applyBoardAccessFilter/queryWorkspaceSearch
+// control flow executes unmodified, while every clause actually applied is logged.
+class FakeQueryBuilder {
+  constructor(
+    private readonly table: string,
+    private readonly rows: FakeRow[],
+  ) {}
+
+  select(..._args: unknown[]): this {
+    calls.push(`${this.table}.select`);
+    return this;
+  }
+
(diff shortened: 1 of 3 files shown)


hermesrobinson-dev (approved):
Independent parent review of 7b4233b41d3b2660a8612808307afbe4e8c3f802: this PR recovers a worker that timed out after the production typing commit but before adding tests -- I fixed 5 lint findings in its uncommitted test file (redundant non-null assertions after assert.ok narrowing), verified, committed and pushed as matthewvryanjh myself. Authorization-sensitive endpoint (board-access role filtering for search) reviewed with extra care: typed board/card search row projections via migration-derived interfaces; applyBoardAccessFilter's qb param now Knex.QueryBuilder instead of ReturnType<typeof db>, no logic changed. Two disclosed no-op deltas confirmed via emitted-JS diff (redundant includeArchived??false and Number(rank) coercions dropped, both no-ops). The new test builds a real recording FakeQueryBuilder that exercises actual applyBoardAccessFilter branching (not a stub) across OWNER/ADMIN/MEMBER/VIEWER/GUEST roles, confirming GUEST never reaches the MEMBER/VIEWER whereIn branch, rank-sorted merge, background proxying, and rank exclusion from the public shape. Reran independently: 1 pass, 3 assertions covering the full scenario set. Focused lint 0 errors on all 3 files. Typecheck byte-identical base/head. Full suite 1253 to 1254 pass, 180 fail/30 errors unchanged. Build:client and diff --check passed. Required verify passed.

PR #300, opened 2026-09-10
fix(e2e): correct list-management spec to real list API contract

matthewvryanjh:
## Summary

Tests 1, 2, 4 in `list-management.spec.ts` used the wrong payload/route for the list API:

- **Create (Test 1):** sent `{ name, position: 0 }`, API requires `{ title }`
- **Rename (Test 2):** sent `{ name }`, API requires `{ title }`
- **Archive (Test 4):** `PATCH /lists/:id` with `{ archived: true }`, but the archive route is `PATCH /lists/:id/archive`

All now match the actual contract (`server/extensions/list/api/*`). Test 5 (unauth guard) passes unchanged.

## Verification

- ESLint clean
- Focused run: Tests 1, 2, 3, 4, 5 all pass (4 passed + Test 5); the 4 skipped are UI tests (6, 7, 8) needing the UI server — separate concern

### DIFF
--- a/tests/e2e/list-management.spec.ts
+++ b/tests/e2e/list-management.spec.ts
@@ -27,12 +27,12 @@ test.describe('List Management', () => {
 
   // ── API: Create ──────────────────────────────────────────────────────────────
 
-  test('Test 1 — Create list returns 201 with id and name', async ({ request }) => {
+  test('Test 1 — Create list returns 201 with id and title', async ({ request }) => {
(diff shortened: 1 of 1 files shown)


hermesrobinson-dev (approved):
Independent review APPROVE. Test-only (1 file). Create/rename now send { title } matching the list API contract; archive uses the correct PATCH /lists/:id/archive route (handler reads no body). Test 5 unauth guard passes (auth runs before body validation). ESLint clean.

PR #301, opened 2026-09-10
fix(e2e): board-members spec uses real users and correct API contract

matthewvryanjh:
## Summary

The `board-members.spec.ts` used fake user IDs (`user-2`/`user-3`, and the JWT as `:userId`) that aren't workspace members, so the API correctly returned 422/404. It also asserted the wrong response field (`userId` vs `user_id`) and error name (`last-admin` vs `last-board-admin`).

## Changes (test-only)

- Register a real second user and add them to the workspace before adding to the board (`POST /workspaces/:id/members` then `POST /boards/:id/members`)
- Use the creator's real `user.id` for the last-ADMIN PATCH/DELETE guards
- Assert `user_id` (GET response shape) and `last-board-admin` error name

## Verification

- All 5 tests pass
- ESLint clean

### DIFF
--- a/tests/e2e/board-members.spec.ts
+++ b/tests/e2e/board-members.spec.ts
@@ -12,7 +12,7 @@ async function registerAndLogin(request: APIRequestContext, suffix: string) {
     data: { email, password, name: `BM ${suffix}` },
   });
   const body = await regRes.json();
-  return { token: body.data.accessToken, email };
+  return { token: body.data.accessToken, userId: body.data.user.id, email };
 }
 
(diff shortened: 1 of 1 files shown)


hermesrobinson-dev (approved):
Independent review APPROVE. Test-only (1 file). Uses real second user (register user.id), adds to workspace then board; asserts user_id (GET shape) and last-board-admin error name. Creator is sole ADMIN so owner.userId correctly triggers the last-ADMIN guard. ESLint clean.

PR #302, opened 2026-09-10
fix(e2e): board-view-preference spec matches real API contract

matthewvryanjh:
## Summary

Two tests in `board-view-preference.spec.ts` didn't match the real API contract:

- **Access-denied test** used a non-existent board (`board123`); the router resolves the board ID before auth, so it returned 404 not 401. Now creates a real board so the unauthenticated request returns 401.
- **Invalid-view-type test** asserted `body.name`, but the API returns `{ error: { code: 'invalid-view-type' } }`. Now asserts `body.error.code`.

## Verification

- All 5 tests pass
- ESLint clean

### DIFF
--- a/tests/e2e/board-view-preference.spec.ts
+++ b/tests/e2e/board-view-preference.spec.ts
@@ -73,11 +73,16 @@ test.describe('Board View Preference API', () => {
   });
 
   test('Access denied without authentication', async ({ request }) => {
+    // Create a real board so the router resolves it; auth runs after board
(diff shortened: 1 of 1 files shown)


hermesrobinson-dev (approved):
Independent review APPROVE. Test-only (1 file). Access-denied now creates a real board (router resolves board ID before auth → 401 on real board, 404 on non-existent); invalid-view-type asserts body.error.code matching the API. ESLint clean.

PR #303, opened 2026-09-10
fix(e2e): otel-metrics 400 tests respect OTEL_ENABLED

matthewvryanjh:
## Summary

The propagation endpoint only validates the body (returns 400) when `OTEL_ENABLED=true`; when OTEL is off it returns 204 regardless (documented contract). The missing/negative `delayMs` tests hard-coded 400, failing under the local default (`OTEL_ENABLED=false`).

## Change (test-only)

Assert 400 when `OTEL_ENABLED=true`, 204 otherwise.

## Verification

- All 6 tests pass
- ESLint clean

### DIFF
--- a/tests/e2e/otel-metrics.spec.ts
+++ b/tests/e2e/otel-metrics.spec.ts
@@ -51,18 +51,22 @@ test.describe('POST /api/v1/metrics/propagation', () => {
     expect(res.status()).toBe(204);
   });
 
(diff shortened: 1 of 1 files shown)


hermesrobinson-dev (approved):
Independent review APPROVE. Test-only (1 file). The propagation endpoint only validates (400) when OTEL_ENABLED=true; fix asserts 400 when on, 204 otherwise, matching the contract. Card-move smoke routes correct. ESLint clean.

PR #304, opened 2026-09-10
fix(e2e): UI tests navigate to UI server and correct board route

matthewvryanjh:
## Summary

`timeline-view` and `calendar-view` used `BASE_URL` (the API server, 3000) for `page.goto`, so the browser hit the API's `/login` (404) and never rendered the login form — every test timed out on `input[type=email]`. They also navigated to `/boards/:id`, but the board detail route is `/b/:id`.

## Changes (test-only)

- Add `UI_URL` (`TEST_UI_URL`, 5173) and use it for page navigation
- Navigate to `/b/:boardId` instead of `/boards/:boardId`

## Verification

- Unblocks timeline-view from 0 to 4 passing
- ESLint clean

### DIFF
--- a/tests/e2e/calendar-view.spec.ts
+++ b/tests/e2e/calendar-view.spec.ts
@@ -11,6 +11,7 @@
 import { test, expect, type APIRequestContext, type Page } from '@playwright/test';
 
 const BASE_URL = process.env.TEST_BASE_URL ?? 'http://localhost:3000';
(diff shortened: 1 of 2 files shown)


hermesrobinson-dev (approved):
Independent review APPROVE. Test-only (2 files). UI tests now navigate to the UI server (5173) via UI_URL and to the correct board route /b/:id (no /boards/:id route exists — falls to NotFound). Login selectors match the app. ESLint clean.

PR #305, opened 2026-09-11
fix(e2e): UI specs navigate to UI server and correct board route

matthewvryanjh:
## Summary

14 UI specs used `BASE_URL` (the API server, 3000) for `page.goto`, so the browser hit the API's `/login` (404) and never rendered the login form — every UI test timed out. They also navigated to `/boards/:id`, but the board detail route is `/b/:id` (no `/boards/:id` route exists; falls to NotFound).

## Changes (test-only)

- Add `UI_URL` (`TEST_UI_URL`, 5173) where missing and use it for page navigation
- Navigate to `/b/:boardId` instead of `/boards/:boardId`

## Verification

- Unblocks timeline-bar from 0 to 4 passing
- ESLint clean on all 14 files

### DIFF
--- a/tests/e2e/attachment-upload.spec.ts
+++ b/tests/e2e/attachment-upload.spec.ts
@@ -211,7 +211,7 @@ test.describe('Attachment Upload', () => {
 
     await page.goto(UI_URL);
     await page.evaluate(({ t }: { t: string }) => localStorage.setItem('auth_token', t), { t: token });
-    await page.goto(`${UI_URL}/boards/${boardId}`);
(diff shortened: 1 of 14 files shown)


hermesrobinson-dev (approved):
Independent review APPROVE (after fix). The table-view window.location.href now uses UI_URL (5173) instead of BASE_URL (3000) — the one REQUEST_CHANGES finding is resolved. All 14 files test-only, UI nav on 5173, board route /b/:id, API calls on 3000. ESLint clean.

PR #306, opened 2026-09-11
fix(e2e): use register token directly to avoid login rate limit

matthewvryanjh:
## Summary

5 specs did a separate `POST /auth/token` login call after register, which hits the login rate limit (10/IP/min) and 429s under the full suite — `body.data` is undefined, so reading `accessToken` throws. Register returns an accessToken directly (201), so the redundant login call is dropped.

## Changes (test-only)

- calendar-view, timeline-view, timeline-drag, timeline-bar, notification-type-prefs: capture token from register response instead of a separate login call

## Verification

- Unblocks calendar-view from 0 to 6 passing
- ESLint clean on all 5 files


hermesrobinson-dev (approved):
Independent review APPROVE. Test-only (5 files). Register returns accessToken on 201 (no rate limit); login is rate-limited 10/IP/min. Capturing token from register response is correct and preserves the Credentials shape. ESLint clean.

PR #307, opened 2026-09-11
fix(e2e): log in via UI form instead of localStorage token injection

matthewvryanjh:
## Summary

The app authenticates via HttpOnly cookies (refresh + access tokens) set by the login/register response — it does NOT persist auth to localStorage. 10 specs injected `auth_token` into localStorage, which the app ignores, so on reload it booted unauthenticated and showed the login page; every UI test then timed out waiting for board elements.

## Changes (test-only)

- Add `registerAndGetCredentials` + `loginViaUi` helpers to `_helpers.ts`
- Update 10 specs to register via API (for the token) then log in through the UI form so the browser receives the HttpOnly cookies

## Verification

- Unblocks board rendering for card-description, custom-fields, table-view, delete-confirmation, list-management, search-filters, attachment-upload, payment-card-price, presence-board
- ESLint clean on all 11 files

### DIFF
--- a/tests/e2e/_helpers.ts
+++ b/tests/e2e/_helpers.ts
@@ -73,3 +73,39 @@ export async function createCard(
   const body = await res.json() as { data: { id: string } };
   return body.data.id;
 }
+
+export interface Credentials {
+  email: string;
+  password: string;
+  token: string;
+}
+
+// Register a user and return credentials (email/password/token). The app
+// authenticates the browser via HttpOnly cookies set by the login/register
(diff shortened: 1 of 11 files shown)


hermesrobinson-dev (approved):
Independent review APPROVE. Test-only (11 files). App authenticates via HttpOnly cookies (no localStorage); loginViaUi correctly logs in through the form so the browser receives the cookies. table-view rewrite correct. ESLint clean.

PR #308, opened 2026-09-11
fix(e2e): inject refresh_token cookie instead of rate-limited UI login

matthewvryanjh:
## Summary

The UI form login hits the login rate limit (10/IP/min) and 429s under the full suite, so `loginViaUi` timed out waiting for `/workspaces`. Register is not rate-limited and sets the HttpOnly `refresh_token` cookie, which the app's boot `refreshTokenThunk` reads to restore auth.

## Changes (test-only)

- `registerAndGetCredentials` now captures the `refresh_token` from the register response's `set-cookie` header
- Add `loginViaCookie` to inject that cookie into the browser context
- 10 specs use `loginViaCookie` instead of `loginViaUi`

## Verification

- Unblocks card-description test 1 (no login timeout)
- ESLint clean on all 11 files

### DIFF
--- a/tests/e2e/_helpers.ts
+++ b/tests/e2e/_helpers.ts
@@ -78,11 +78,13 @@ export interface Credentials {
   email: string;
   password: string;
   token: string;
+  refreshToken: string;
 }
 
-// Register a user and return credentials (email/password/token). The app
-// authenticates the browser via HttpOnly cookies set by the login/register
-// response, so UI specs must log in through the form (not localStorage).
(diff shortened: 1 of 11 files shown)


hermesrobinson-dev (approved):
Independent review APPROVE. Test-only (11 files). Register sets the HttpOnly refresh_token cookie (no rate limit); boot refreshTokenThunk restores auth from it. loginViaCookie injects the cookie matching Path=/api/v1/auth/refresh. API calls unchanged. ESLint clean.

PR #309, opened 2026-09-11
fix(e2e): inject refresh_token cookie in timeline/calendar/offline specs

matthewvryanjh:
## Summary

6 specs (timeline-view, calendar-view, timeline-drag, timeline-bar, notification-type-prefs, offline-queue-persistence) used their own rate-limited UI form login (`goToBoard`/`loginViaUi`), which 429s under the full suite and times out waiting for `/workspaces`. Register is not rate-limited and sets the HttpOnly `refresh_token` cookie, which the app's boot `refreshTokenThunk` reads to restore auth.

## Changes (test-only)

- Capture `refresh_token` from the register response's `set-cookie` header
- Inject it into the browser context instead of the UI form login

## Verification

- Unblocks timeline-view from 4 to 5 passing (no login timeouts)
- ESLint clean on all 6 files

### DIFF
--- a/tests/e2e/calendar-view.spec.ts
+++ b/tests/e2e/calendar-view.spec.ts
@@ -15,7 +15,7 @@ const UI_URL = process.env.TEST_UI_URL ?? 'http://localhost:5173';
 
 // ── Helpers ────────────────────────────────────────────────────────────────
 
-interface Credentials { email: string; password: string; token: string }
+interface Credentials { email: string; password: string; token: string; refreshToken: string }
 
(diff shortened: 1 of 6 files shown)


hermesrobinson-dev (approved):
Independent review APPROVE. Test-only (6 files). Register sets the HttpOnly refresh_token cookie (no rate limit); boot refreshTokenThunk restores auth from it. Cookie injection matches Path=/api/v1/auth/refresh. ESLint clean.

PR #310, opened 2026-09-11
fix(e2e): card-tile and card-modal selectors match actual DOM

matthewvryanjh:
## Summary

`CardTile` renders as `div[role=button][aria-label^="Card:"]` with no card-tile testid, so the old selector timed out. The card modal has `aria-label="Card: <title>"` and a hidden mobile sidebar also uses `role="dialog"`, so the generic `[role=dialog]` selector matched the wrong element first.

## Changes (test-only)

- Use `[aria-label^="Card:"]` to open the card modal
- Target the modal as `[role="dialog"][aria-label^="Card:"]` to avoid the hidden mobile sidebar

## Verification

- Both specs (card-description-inline-edit, custom-fields-ui) now pass the card-tile and modal-open steps (previously timed out on the selector)
- ESLint clean on both files

### DIFF
--- a/tests/e2e/card-description-inline-edit.spec.ts
+++ b/tests/e2e/card-description-inline-edit.spec.ts
@@ -35,8 +35,8 @@ test.describe('Card Description — Inline Click-to-Edit', () => {
     await page.waitForLoadState('networkidle');
 
     // Open the card modal
-    await page.locator('[data-testid="card-tile"], .card-tile, [class*="card"]').first().click();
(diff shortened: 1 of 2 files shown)


hermesrobinson-dev (approved):
Independent review APPROVE. Test-only (2 files). CardTile is div[role=button][aria-label^=Card:] (no card-tile testid); card modal is role=dialog[aria-label^=Card:] (bare [role=dialog] matches hidden mobile sidebar). Selectors now correct. ESLint clean.

PR #312, opened 2026-09-11
fix(e2e): recover custom fields UI spec

matthewvryanjh:
## Summary

Fixes `tests/e2e/custom-fields-ui.spec.ts` after the card tile/modal selector drift was unblocked.

## Root causes

- `SetupResult` did not expose `token`, but tests destructured `const { token, boardId } = ...`, so custom-field API creation sent `Authorization: Bearer undefined` and failed at `body.data.id`.
- The spec loaded the browser board before creating each custom field, then reloaded. Because app boot refresh rotates auth cookies and access state is in memory, that reload could land on `/login`.
- A few assertions/selectors were too broad after badges rendered, e.g. `getByText('8')` matching timestamps and `getByLabel(/Notes/i)` matching the badge, input, and clear button.
- Checkbox save is async; the persisted value is reflected by the rendered `Reviewed: ✓` badge.

## Changes (test-only)

- Return `token` from setup fixture.
- Create custom fields before opening the UI board; then inject cookie and navigate once.
- Add explicit custom-field API failure diagnostics.
- Use exact accessible field controls/badge labels:
  - `Notes value`
  - `Story Points value` / `Story Points: 8`
  - `Internal Notes value`
  - `Reviewed: ✓`
- Preserve product behaviour; no source changes.

## Verification

- `bunx playwright test tests/e2e/custom-fields-ui.spec.ts --project=e2e --reporter=line`: 6 passed (10.2s)
- `bunx eslint tests/e2e/custom-fields-ui.spec.ts`: clean
- `git diff --check`: clean

### DIFF
--- a/tests/e2e/custom-fields-ui.spec.ts
+++ b/tests/e2e/custom-fields-ui.spec.ts
@@ -10,23 +10,26 @@ const UI_URL = process.env.TEST_UI_URL ?? 'http://localhost:5173';
 
 interface SetupResult {
   creds: Credentials;
+  token: string;
   boardId: string;
   cardId: string;
 }
 
-async function setupBoardWithCard(request: APIRequestContext, page: import('@playwright/test').Page): Promise<SetupResult> {
+async function setupBoardWithCard(request: APIRequestContext): Promise<SetupResult> {
   const creds = await registerAndGetCredentials(request, 'cf-ui');
   const wsId = await createWorkspace(request, creds.token);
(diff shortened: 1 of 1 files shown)


hermesrobinson-dev (approved):
Independent review APPROVE. Test-only. Fixes missing token, avoids UI reload/auth rotation drift, tightens selectors to actual accessible names/badges, and verifies checkbox persistence via Reviewed badge. ESLint clean; targeted spec 6/6 passed.

PR #313, opened 2026-09-11
fix(e2e): recover custom fields board settings panel spec

matthewvryanjh:
## Summary

Recovers `tests/e2e/custom-fields-board-panel.spec.ts` (0/7 -> 7/7). Part of the full-suite e2e recovery.

## Root causes

- **Panel never opened:** the board header renders a `Board settings` menu *trigger* (lowercase). The settings panel (`aria-label="Board Settings"`) only mounts after clicking the menu item. Tests clicked once and timed out.
- **Missing token:** `setupBoardAndNavigate` returned `{ creds, boardId }` but tests destructured `token`, so seeded custom-field API calls sent `Bearer undefined` and no fields existed.
- **DOM contract drift:** labels are `Field name`, `Field type`, `Rename field <name>`, `Delete field <name>`, `Show on card tile`, `Label for dropdown option`, `+ Add option`, `New custom field form`, `Dropdown options editor`, `Create custom field`.
- **Navigation race:** a redundant second `goto` after seeding raced boot token refresh and landed on `/workspaces`.
- **Dialog ordering:** `window.confirm` handler was registered *after* the delete click.
- **Controlled checkbox:** `check()` asserts synchronously; state updates only after the API refetch, so the assertion must auto-retry.

## Changes (test-only)

- `openBoardSettings` helper drives trigger + menu item + panel-visible assertion.
- `setupBoardAndNavigate` returns `token` and retries board navigation until the header renders.
- Selectors aligned to the real accessible names listed above.
- Dialog handler registered before delete; controlled checkbox asserted via click + retrying `toBeChecked`.

## Verification

- `bunx playwright test tests/e2e/custom-fields-board-panel.spec.ts --project=e2e`: **7 passed** twice consecutively (10.8s, 10.0s)
- `bunx eslint tests/e2e/custom-fields-board-panel.spec.ts`: clean
- `git diff --check`: clean
- No product source changed.

### DIFF
--- a/tests/e2e/custom-fields-board-panel.spec.ts
+++ b/tests/e2e/custom-fields-board-panel.spec.ts
@@ -11,40 +11,56 @@ const UI_URL = process.env.TEST_UI_URL ?? 'http://localhost:5173';
 async function setupBoardAndNavigate(
   request: APIRequestContext,
   page: import('@playwright/test').Page,
-): Promise<{ creds: Credentials; boardId: string }> {
+): Promise<{ creds: Credentials; token: string; boardId: string }> {
   const creds = await registerAndGetCredentials(request, 'cf-panel');
   const wsId = await createWorkspace(request, creds.token);
   const boardId = await createBoard(request, creds.token, wsId);
 
   // The app authenticates via HttpOnly cookies, so log in through the UI form.
   await loginViaCookie(page, UI_URL, creds);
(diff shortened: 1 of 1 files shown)


hermesrobinson-dev (approved):
Independent review APPROVE. Test-only. Two-step Board settings flow, real accessible names, dialog handler ordering, and controlled checkbox assertion all verified against source. ESLint clean; spec ran 7 passed twice.

PR #315, opened 2026-09-11
fix(e2e): recover timeline view and bar specs

matthewvryanjh:
## Summary

Recovers `tests/e2e/timeline-view.spec.ts` and `tests/e2e/timeline-bar.spec.ts` (8/17 -> 13/13 across the two specs). Test-only.

## Root causes

- **Boot-refresh navigation race:** navigating right after cookie injection sometimes landed on `/workspaces`; navigation now retries until `board-view-switcher` renders.
- **Deliberate product change (eb510421, "improve card filtering logic"):** `TimelineView` now sets `unscheduledCards: []` — cards without a due_date are not displayed at all, and a card with only a due_date is scheduled (start_date defaults to today). Three tests still asserted the removed unscheduled-chip behaviour and one asserted a no-date card renders as a chip. Rewritten to assert the current contract:
  - due_date-only card renders as a scheduled bar
  - no-date card is not rendered (no bar, no chip)
  - clicking a scheduled bar opens the card modal
- **Card modal URL:** the app opens the card via its short link, so `?card=<short-id>`; the test asserted the UUID. Now asserts `?card=` presence.

## Not included (deliberate)

`tests/e2e/timeline-drag.spec.ts` is left untouched. Its 3 failures are the same root cause as the parked calendar due-date bug: the server stores `start_date`/`due_date` shifted to server-local midnight (UTC-stored), so a client round-trip loses a day. That is a product/data issue, not test drift, and is tracked in the recovery backlog rather than masked here.

## Verification

- `bunx playwright test tests/e2e/timeline-view.spec.ts tests/e2e/timeline-bar.spec.ts --project=e2e`: **13 passed** twice consecutively (13.6s, 11.3s)
- `bunx eslint` on both files: clean
- `git diff --check`: clean
- No product source changed.

### DIFF
--- a/tests/e2e/timeline-bar.spec.ts
+++ b/tests/e2e/timeline-bar.spec.ts
@@ -80,14 +80,24 @@ function offsetDate(offsetDays: number): string {
   return `${d.getFullYear()}-${String(d.getMonth() + 1).padStart(2, '0')}-${String(d.getDate()).padStart(2, '0')}`;
 }
 
+// App boot performs an async token refresh; navigating immediately can race and
+// land on /workspaces. Retry until the board view switcher renders.
+async function gotoBoardUntilReady(page: Page, baseUrl: string, boardId: string): Promise<void> {
+  for (let attempt = 0; attempt < 3; attempt++) {
+    await page.goto(`${baseUrl}/b/${boardId}`);
+    await page.waitForLoadState('networkidle');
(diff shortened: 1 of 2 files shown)


hermesrobinson-dev (approved):
Independent review APPROVE. Test-only (2 specs). TimelineView unscheduledCards:[] product change and ?card=<shortId> short-link URL both verified against source. ESLint clean; 13 passed twice independently.

PR #318, opened 2026-09-11
fix(e2e): recover five singleton specs

matthewvryanjh:
## Summary

Recovers five singleton-failure specs (7/20 -> 14/14 passing, 6 legitimately skipped). Test-only.

### business-logic-invariants
Board archiving is `PATCH /boards/:id/archive` (toggles ACTIVE <-> ARCHIVED). The spec sent `PATCH /boards/:id { archived: true }`, which returns 400, so the board was never archived and the read-only assertions were meaningless.

### mcp-http-init
MCP streamable HTTP requires the client to accept both JSON and SSE. Without `Accept: application/json, text/event-stream` the server returns 406 (verified: no header -> 406, both -> 200, SSE-only -> 406).

### presence-board
Presence is populated by the realtime WebSocket layer (`presence:<boardId>:<userId>` cache keys set on subscribe), **not** by REST. There is no `POST`/`DELETE /boards/:id/presence` (both 404), which is why tests 2 and 3 skip. Test 4 was asserting a REST-populated list and always got `[]`. It now subscribes two users over `/api/v1/ws`, then reads the REST presence list and expects both. User B is added to the workspace first because `subscribe` enforces workspace membership.

### notification-type-prefs
- `switches.first()` matched the **master** "Toggle notifications for this board" switch; clicking it disabled the entire section including the Reset button, so the later click timed out. Locators are now scoped to the per-type matrix `table`.
- The board-override indicator is `ring-indigo-400`. Matching the substring `ring-indigo` also matched the always-present `focus-visible:ring-indigo-500` utility, so the "no override after reset" assertion could never pass. Tightened to `ring-indigo-400`.

### table-view
Boot-refresh navigation race: navigating immediately after cookie injection sometimes landed on `/workspaces`. Navigation now retries until the view switcher renders.

## Verification

- `bunx playwright test` on all five specs: **14 passed, 0 failed** across three consecutive runs (6.9s, 6.6s, 8.5s)
- `bunx eslint` on all five files: clean
- `bunx tsc --noEmit`: 146 errors, unchanged (none in touched files)
- `git diff --check`: clean
- No product source changed.

## Note

The presence fix works around a genuine product inconsistency: workspace member-add lowercases the email before lookup while `register` preserves case, so any mixed-case address cannot be matched. Documented in the recovery backlog, not masked.

### DIFF
--- a/tests/e2e/business-logic-invariants.spec.ts
+++ b/tests/e2e/business-logic-invariants.spec.ts
@@ -21,10 +21,10 @@ test.describe('Business Logic Invariants', () => {
       listId = await createList(request, token, boardId);
       cardId = await createCard(request, token, listId, 'Test Card');
 
-      // Archive the board
-      const archiveRes = await request.patch(`${BASE_URL}/api/v1/boards/${boardId}`, {
+      // Archive the board — the archive endpoint is PATCH /boards/:id/archive
+      // and toggles ACTIVE <-> ARCHIVED.
+      const archiveRes = await request.patch(`${BASE_URL}/api/v1/boards/${boardId}/archive`, {
         headers: { Authorization: `Bearer ${token}` },
-        data: { archived: true },
       });
       expect(archiveRes.status()).toBeLessThan(300);
     });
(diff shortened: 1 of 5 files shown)


hermesrobinson-dev (approved):
Independent review APPROVE. Test-only (5 specs). Archive endpoint, MCP 406/Accept, WS-driven presence, ring-indigo-400 precision, and table-scoped switches all verified against source and live behaviour. ESLint clean; 14 passed / 0 failed independently.

PR #320, opened 2026-09-11
fix(e2e): attach replay request listener before reload

matthewvryanjh:
## Summary

Fixes a race that only surfaced in the full suite (the second offline-queue test passed in isolation but timed out at 15s in the suite).

## Root cause

`page.waitForRequest(...)` for the replay POST was registered **after** `page.reload()`. But the reload itself triggers boot hydration and the WS-connect replay, so the listener could miss the request entirely and time out. Timing differs between an isolated run and the full suite, which is why it passed standalone.

## Fix

Attach the listener before reloading, and drop the redundant second `goto` (the reload already boots the app and connects the WS, so replay fires from it).

## Verification

- `bunx playwright test tests/e2e/offline-queue-persistence.spec.ts --project=e2e`: **2 passed** three times consecutively (4.3s, 3.7s, 3.3s)
- `bunx eslint`: clean
- `git diff --check`: clean
- No product source changed.

### DIFF
--- a/tests/e2e/offline-queue-persistence.spec.ts
+++ b/tests/e2e/offline-queue-persistence.spec.ts
@@ -207,19 +207,18 @@ test.describe('offline mutation queue — IndexedDB persistence', () => {
       enqueuedAt: Date.now(),
     };
 
-    // Seed mutation and reload
+    // Seed mutation, then reload. The listener must be attached BEFORE the
+    // reload because boot hydrates the queue and replay fires as soon as the WS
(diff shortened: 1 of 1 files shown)


hermesrobinson-dev (approved):
Independent review APPROVE. Test-only (1 file, +4/-5). Boot hydrate -> WS replay race verified against source; listener now attached before reload, redundant goto removed, both assertions retained. ESLint clean; 2 passed twice in isolation.

PR #321, opened 2026-09-11
fix(dates): interpret card date fields in the viewer local timezone

matthewvryanjh:
## Summary

Fixes the due-date timezone inconsistency (the parked product bug). Card date fields are now interpreted in the viewer's local timezone on every surface, so the calendar and timeline agree with the card tile / meta strip / table. Also unblocks the acceptance gate.

## Root cause

`due_date` and `start_date` are stored as ISO **instants**, but surfaces disagreed on how to read them:

| Surface | Rendered |
|---|---|
| CardItem due badge | local date + time |
| CardMetaStrip | local date + time |
| CardTile | local date |
| TableRow | local date |
| **CalendarView** | **raw UTC slice** (`iso.slice(0, 10)`) |
| **TimelineView** | **raw UTC slice** |

On a `+07` server, `due_date: "2026-09-15"` is stored as `2026-09-14T17:00:00Z`, so Calendar and Timeline showed the card on the **14th** while the card surfaces showed the **15th**.

Note: these fields are genuinely instants — the date picker has a time input (`H:MM AM`, defaults 12:00 PM) and persists it. So the fix is *not* to convert them to date-only; it is to interpret them consistently in local time. (I initially proposed date-only normalisation; that would have destroyed user-set times and hidden them in the UI. Corrected before writing code.)

## Changes

New `src/common/utils/dates.ts`: `localDateKey`, `parseLocalDate`, `toLocalDateKey`.

Applied everywhere a date field is keyed, grouped, compared, or used for day arithmetic:

- `CalendarView` — `buildCardsByDay`
- `useCalendarDrag` — no-op comparison
- `TimelineBar`, `TimelineRow`, `useTimelineDrag` — day math and drag originals
- `TimelineView` — `todayIso` (was `toISOString().slice(0,10)`, i.e. the UTC date) and the `scheduledCards` filter

Test-side blockers fixed so the acceptance gate runs reliably:

- Retry board navigation until it renders (the boot-refresh race), in `calendar-view`, `timeline-drag`, `custom-fields-ui`.
- Integer mouse-move `steps` (`10.5` was rejected by Playwright).
- Compare server dates on the local calendar date rather than the UTC slice.

## Verification

- `calendar-view` + `timeline-drag` + `timeline-view` + `timeline-bar`: **27 passed** twice consecutively (19.4s, 15.4s)
- Full suite: **156 passed / 0 failed / 20 skipped** (previously 150/17/20 — the 7 parked calendar/timeline failures now pass, and the 3 payment failures are excluded by standing rule; attachments remain parked)
- `bunx eslint`: no new findings (5 pre-existing errors in these files, unchanged)
- `bunx tsc --noEmit`: 146 errors, unchanged
- `git diff --check`: clean

### DIFF
--- a/src/common/utils/dates.test.ts
+++ b/src/common/utils/dates.test.ts
@@ -1,0 +1,56 @@
+import { describe, expect, it } from 'bun:test';
+import { localDateKey, parseLocalDate, toLocalDateKey } from './dates';
+
+// These tests guard the two date traps documented in dates.ts:
+//   1. slicing an ISO string yields the UTC date, not the local one;
+//   2. new Date("YYYY-MM-DD") is UTC midnight, so converting it to local time
+//      shifts the day for negative-offset timezones.
+// A date-only string must always survive as the same calendar date, regardless
+// of the runtime timezone, because drag/resize paths persist and re-read it.
+
+describe('localDateKey', () => {
+  it('returns a date-only string verbatim', () => {
+    expect(localDateKey('2026-09-15')).toBe('2026-09-15');
+    expect(localDateKey('2026-01-01')).toBe('2026-01-01');
+    expect(localDateKey('2026-12-31')).toBe('2026-12-31');
+  });
+
+  it('does not shift date-only strings across month and year boundaries', () => {
+    expect(localDateKey('2026-01-01')).toBe('2026-01-01');
+    expect(localDateKey('2026-03-01')).toBe('2026-03-01');
+    expect(localDateKey('2026-12-01')).toBe('2026-12-01');
+  });
+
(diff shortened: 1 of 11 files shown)


hermesrobinson-dev (approved):
Independent review APPROVE (head 696f40ea; commit 2 verified line-by-line). Date-only strings now return verbatim (no new Date round-trip), so the west-of-UTC regression is genuinely fixed; unit tests pass under both TZ=America/Los_Angeles and TZ=Asia/Bangkok. No raw slice remains in Calendar/Timeline. ESLint identical to main (no new errors); tsc unchanged. No server/migration change; user-set times preserved.

PR #322, opened 2026-09-11
fix(storage): make FLAG_USE_LOCAL_STORAGE work and unblock attachment tests

matthewvryanjh:
## Summary

Implements Option A: make `FLAG_USE_LOCAL_STORAGE` actually work so attachment tests and local dev can run without AWS.

`FLAG_USE_LOCAL_STORAGE` was documented in `.env.example` ("When `FLAG_USE_LOCAL_STORAGE=true` (default in dev), these are auto-set to LocalStack") and declared in `server/mods/flags/defaults.ts` — but **nothing read it**. With `S3_ENDPOINT` empty and no credentials, every S3 call failed: `multipart/start` returned 502 and the attachment e2e specs could not run.

## Changes

**`server/config/env.ts`** — add `resolveS3Env()`. With the flag on it defaults the endpoint to LocalStack (`http://localhost:4566`, overridable via `LOCALSTACK_ENDPOINT`) and fills in LocalStack's throwaway credentials. Explicit env values always win, so MinIO or real credentials still work with the flag on. Empty strings count as unset, because `.env` ships `S3_ENDPOINT=` and `""` does not fall through a nullish coalesce.

**`server/extensions/attachment/common/config/s3.ts`** — set `requestChecksumCalculation` / `responseChecksumValidation` to `WHEN_REQUIRED`. `@aws-sdk` 3.1000 adds a default CRC32 checksum that S3-compatible stores (LocalStack 3.x, older MinIO) reject with `Checksum Type mismatch`, which broke presigned PUTs and multipart parts.

**`server/extensions/attachment/api/multipart/complete.ts`** — accept the AWS wire format `PartNumber`/`ETag` alongside camelCase. Only the ETag tolerated both casings; the part number did not, so a client sending `PartNumber` (the AWS-correct shape) produced `PartNumber: undefined` and S3 returned `InternalError`. This is a real product bug, not test drift — an AWS SDK client would hit it.

**`docker-compose.yml` / `docker-compose.prod.yml`** — pin localstack to `3.8.1`. The `latest` tag now requires a licence token and exits 55, so the documented `docker compose up -d localstack` was broken on a fresh pull.

**`tests/e2e/attachment-upload.spec.ts`** — do a real part upload when storage is reachable instead of stubbing an ETag that S3 cannot complete; assert the fields the confirm endpoint actually returns (`mime_type`, `s3_key`) rather than `content_type`/`url`, which only `serializeAttachment` produces.

## Verification

Full multipart flow against LocalStack — create → part-url → real 5 MiB PUT → complete → confirm — returns `status: READY`, `mime_type: image/png`.

- `attachment-upload.spec.ts`: **6 passed / 0 failed / 2 skipped**, twice consecutively (was 2 failing)
- Full suite: **158 passed / 9 failed / 20 skipped** (was 156/11/20)
- ESLint clean on touched files; no new tsc errors (146 baseline unchanged)

## Not in scope

- `attachments/attachmentPanel.spec.ts` (5 failures) is still abandoned scaffolding — fake board id, non-existent testids, `expect(true).toBe(true)`. It needs a rewrite with real seeding, which is its own PR.
- `payment-card-price` (3) remains excluded per project rules.

## Local dev note

`docker compose up -d postgres localstack` plus `FLAG_USE_LOCAL_STORAGE=true` is now a working local storage setup. On this machine the compose v2 plugin is unavailable, so `docker-compose` (hyphen) was used.

### DIFF
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -51,7 +51,9 @@ services:
 
   localstack:
     profiles: [local-s3]
-    image: localstack/localstack:latest
+    # Pin to a community (non-pro) tag: the latest tag now requires a licence
+    # token and exits 55. Keep in step with docker-compose.yml.
+    image: localstack/localstack:3.8.1
     ports:
       - "6401:4566"
     environment:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -39,7 +39,10 @@ services:
       retries: 10
 
   localstack:
-    image: localstack/localstack:latest
+    # Pin to a community (non-pro) tag: `latest` now requires a licence token
+    # and exits 55 without one. 3.8.1 is the last freely usable line and is
+    # sufficient for S3. See docs/local-development.md.
+    image: localstack/localstack:3.8.1
     ports:
       - "4566:4566"
     environment:
--- a/server/config/env.ts
+++ b/server/config/env.ts
@@ -10,29 +10,80 @@ function decodeKey(raw: string): string {
   return Buffer.from(raw, 'base64').toString('utf-8');
 }
 
+/**
+ * Resolve the S3 / object-storage settings.
+ *
+ * When FLAG_USE_LOCAL_STORAGE=true we point S3 at the LocalStack service that
+ * docker-compose.yml provides (port 4566) and supply LocalStack's throwaway
+ * credentials, so local dev and e2e runs work with no AWS account. This is the
+ * behaviour .env.example has always documented.
+ *
+ * Any explicit env value wins over the LocalStack default, so a self-hosted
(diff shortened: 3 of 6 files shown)


hermesrobinson-dev (approved):
Independent review APPROVE (head 9b55ece7). Production safety property verified line-by-line: with FLAG_USE_LOCAL_STORAGE unset/false and no explicit values, resolveS3Env() returns exactly the prior values (S3_ENDPOINT '', credentials '', bucket 'kanban', region 'us-east-1'), so flag=false behaviour is unchanged. Empty-string handling via orUnset() is correct. Checksum options and PartNumber/ETag normalisation verified. No new ESLint errors.

PR #323, opened 2026-09-11
fix(seeds): honour FLAG_USE_LOCAL_STORAGE in the Trello import

matthewvryanjh:
## Summary

Follow-up to #322. `db/seeds/trello-import.ts` resolved S3 config on its own and only reacted to `S3_ENDPOINT`:

```ts
const S3_ENDPOINT = Bun.env['S3_ENDPOINT'] || undefined;
```

So a developer with `FLAG_USE_LOCAL_STORAGE=true` still had `S3_ENDPOINT` empty, and the seed script would write seed attachments to **real AWS** rather than LocalStack. Same class of bug as #322, in a second place.

It now applies the same rule as `server/config/env.ts`: the flag defaults the endpoint to `http://localhost:4566`, and an explicit `S3_ENDPOINT` still wins.

## Verification

| Case | Resolved endpoint |
|---|---|
| flag on, no endpoint | `http://localhost:4566` |
| flag off (production) | `undefined` — unchanged, real AWS |
| flag on, explicit `S3_ENDPOINT` | that endpoint |

ESLint unchanged at 5 pre-existing errors; `tsc` unchanged at 146.

Found by the independent review of #322, which noticed the second config path while grepping for `FLAG_USE_LOCAL_STORAGE`.

### DIFF
--- a/db/seeds/trello-import.ts
+++ b/db/seeds/trello-import.ts
@@ -51,7 +51,11 @@ const DATABASE_URL =
 
 const S3_BUCKET    = Bun.env['S3_BUCKET'] ?? 'chimedeck';
 const S3_REGION    = Bun.env['S3_REGION'] ?? 'us-east-1';
-const S3_ENDPOINT  = Bun.env['S3_ENDPOINT'] || undefined;
+// Honour FLAG_USE_LOCAL_STORAGE the same way server/config/env.ts does, so a
+// dev with the flag on does not accidentally write seed attachments to real
+// AWS. An explicit S3_ENDPOINT still wins.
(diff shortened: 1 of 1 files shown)


hermesrobinson-dev (approved):
Independent review APPROVE (head 106f1891). Production safety verified: with FLAG_USE_LOCAL_STORAGE unset/false the expression resolves to undefined, so S3_BASE_URL still targets real AWS. Semantics match server/config/env.ts:36-49 (flag === 'true', explicit S3_ENDPOINT wins); credential fallback correctly yields test/test with the flag on. No new ESLint errors (same 5 pre-existing). Non-blocking nit noted: the seed hardcodes localhost:4566 and ignores a LOCALSTACK_ENDPOINT override — defaults match, so no impact on the goal.

PR #324, opened 2026-09-12
chore(tests): remove the abandoned attachment panel scaffolding

matthewvryanjh:
## Summary

Removes `attachmentPanel.spec.ts` and `attachmentPanel.md`. Agreed with Matt: delete rather than rewrite.

The file is a stub, not a working test. Its own header says *"These tests assume a seeded board and card exist. Adjust selectors/IDs as needed."* and it ends with *"... More tests would be implemented similarly"*.

It cannot pass:

- navigates to `/board/test-board-id` — wrong route shape (real: `/b/:boardId`) and a fake id
- waits on `data-testid` values that appear nowhere in `src/`: `card-title`, `card-modal`, `attachment-panel`, `attach-file-button`, `attachment-list`, `upload-progress-bar`, `delete-attachment-button`
- the clipboard test is a no-op: `expect(true).toBe(true)`
- references `tests/e2e/attachments/fixtures/*.{png,pdf,bin}` — that directory does not exist

Every test either times out in `beforeEach` or asserts nothing.

## Coverage

The real attachment coverage is `tests/e2e/attachment-upload.spec.ts` — 8 tests covering the multipart flow, URL attachments, MIME and size rejection, auth, and both UI display paths. All passing.

`attachmentPanel.md` described 13 scenarios using the same non-existent testids; it goes with the stub.

## Verification

- Full suite: **158 passed / 4 failed / 20 skipped** (was 158 / 9 / 20)
- The five `attachmentPanel` failures are gone; no other spec changed
- No tracked references to the deleted file outside `specs/changelog/` history

## Note on the local BDD map

`specs/bdd/` and `scripts/bdd/` are untracked local artefacts (not in git), but the local `bdd-playwright-map.json` referenced these 5 tests and `validate-bdd.mjs` requires every mapped file to exist. They were regenerated locally so validation no longer reports `attachmentPanel`. That state is local-only and not part of this PR.


hermesrobinson-dev (approved):
Independent review APPROVE (head 3e6ded87). Verified the deleted spec could not pass: wrong route (/board/test-board-id vs real /b/:boardId/:slug?), testids 'attachment-panel'/'upload-progress-bar'/'delete-attachment-button' absent from src/ and tests/, a no-op expect(true).toBe(true), and a non-existent fixtures dir. Coverage check: attachment-upload.spec.ts (8 tests) covers multipart, URL attachments, MIME/size rejection, auth and both UI display paths; the deleted tests contributed zero passing assertions. No tracked references outside changelog history; playwright.config.ts uses testMatch globs so a missing file cannot break it.

PR #325, opened 2026-09-12
fix(e2e): make the attachment UI tests actually run

matthewvryanjh:
## Summary

Tests 7 and 8 of `attachment-upload.spec.ts` never asserted anything. Both opened the board and then soft-skipped with *"Card element not found"*, so the suite reported **2 skipped** and the attachment UI paths (thumbnail, download link) were never exercised.

The independent review of #324 flagged this: *"attachment UI behaviour is thinly exercised overall."* With local storage working (#322) these tests can now run for real.

## Root causes (found by probing the live DOM)

- **Card tiles** were located by `[data-card-id]` / `[data-testid="card-<id>"]` — neither exists. Tiles render as `role=button` with accessible name `Card: <title>`; the other recovered specs use `[aria-label^="Card:"]`.
- **Boot refresh race**: navigating straight after login landed on `/workspaces` or `/login`. Added a `gotoBoard()` retry that waits for the board to render.
- **Stale credentials**: both UI tests used the suite-level `beforeAll` cookie. The app rotates the refresh token on boot, so a later test lands on `/login`. Each UI test now registers its own user, creates its own board/list/card with that user's token, and logs in as that user — a board is only visible to its owner.
- **Attachment section** was located by `[data-testid="attachments"]` / `[aria-label*="attachment"]` — neither exists. The modal renders an `Attachments` heading, an `Attach file` button (`data-testid="attach-file-button"`), and URL attachments as anchors under `Links`.
- **`getByRole('button', {name: 'Attach file'})`** was a strict-mode violation — two controls share that name. Scoped to the stable testid.
- **`getByText('thumb.png')`** matched both the attachment link and its activity-feed entry. Scoped to the link role.
- Removed the soft-skips that were masking all of the above.

## Verification

- `attachment-upload.spec.ts`: **8 passed / 0 failed / 0 skipped**, three consecutive runs (was 6 passed / 2 skipped)
- Full suite: **159 passed / 5 failed / 18 skipped** — two fewer skips, one more passing test
- ESLint clean on the touched file; no new tsc errors

Remaining failures are the three excluded `payment-card-price` tests plus two load-related flakes, unchanged by this PR.

No product code changed — this is a test-only correction. The selectors were guessing; the assertions now match the DOM the app actually renders.

### DIFF
--- a/tests/e2e/attachment-upload.spec.ts
+++ b/tests/e2e/attachment-upload.spec.ts
@@ -281,28 +281,33 @@ test.describe('Attachment Upload', () => {
     const apiHeaders = { Authorization: `Bearer ${uiToken}`, 'Content-Type': 'application/json' };
     const started = await (await request.post(
       `${BASE_URL}/api/v1/cards/${uiCardId}/attachments/multipart/start`,
-      { headers: apiHeaders, data: { filename: 'thumb.png', mimeType: 'image/png', sizeBytes: 1024 * 1024 } },
+      { headers: apiHeaders, data: { filename: 'thumb.png', mimeType: 'image/png', sizeBytes: 5 * 1024 * 1024 } },
     )).json() as { data: { attachmentId: string; uploadId: string; key: string } };
     const { attachmentId, uploadId, key } = started.data;
 
-    const partUrlRes = await (await request.post(
+    const partUrlRes = await request.post(
       `${BASE_URL}/api/v1/cards/${uiCardId}/attachments/multipart/part-url`,
       { headers: apiHeaders, data: { attachmentId, uploadId, key, partNumber: 1 } },
-    )).json() as { data: { url: string } };
+    );
+    expect(partUrlRes.status()).toBe(200);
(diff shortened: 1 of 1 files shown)


hermesrobinson-dev (approved):
Independent reviews APPROVE (head 9f9615b8). Commit 1: verified every DOM contract against src/ (attach-file-button testid, aria-label^='Card:' tiles, modal selector, Attachments heading, link rendering); no masking, assertions strictly stronger than origin/main, per-test credential setup sound. Commit 2: confirmed Test 7 now asserts a real inline image preview via the attachment view proxy (AttachmentItem imagePreviewSrc), the multipart sequence is correct (start -> part-url -> 5 MiB PUT -> complete -> confirm, ETag casing accepted), and ESLint clean. All four commit-2 review findings fixed in commit 3: duplicate 7b removed, sizeBytes made consistent, multipart statuses asserted, img comment corrected. Verified 8 passed / 0 failed / 0 skipped x3; full suite 160 passed / 4 failed / 18 skipped.

PR #326, opened 2026-09-12
fix(e2e): remove dead-endpoint guards in two specs (9 tests never ran)

matthewvryanjh:
## Summary

Nine tests across two specs had **never executed**. Each guarded itself with:

```ts
if (res.status() === 404 || res.status() === 501) {
  test.skip(true, '… endpoint not yet implemented');
  return;
}
```

…and then called a route that does not exist. The guard can never fail, so when the real routes shipped under different shapes the tests kept skipping silently. They reported as *skipped*, which read as fine in every green suite run.

Found by asking "which 18 tests are skipped?" — the `line` reporter prints only a count.

## business-logic-invariants — 3 skipped → all pass

- `GET /api/v1/me` 404s. The route is `/api/v1/users/me`. The `beforeAll` fetch swallowed the failure and left `userId` unset, which made three tests skip. It now **throws on a non-200**, so a wrong route fails loudly instead of silently skipping.
- Removed the three now-unreachable `if (!userId) test.skip(...)` guards.
- The invite payload sent `{ userId, role }`; `POST /workspaces/:id/members` takes the user's **email**. Switched to `registerAndGetCredentials` to obtain it.
- Tightened a weak assertion that accepted `[200, 204, 422]` for an operation its own title says must succeed. It now requires 2xx and verifies an OWNER remains.

## search-filters — 6 skipped → all pass

- `GET /api/v1/search` does not exist. Real routes: `GET /api/v1/workspaces/:id/search` and `GET /api/v1/boards/:id/search`.
- Test 3's board scoping used an ignored `?boardId=` query param; it now uses the board route.
- Tests 5/6 looked for a board-page search input that does not exist. Search is a **modal palette** opened from the header Search button, with input placeholder `Search boards and cards…` and All/Boards/Cards tabs (`role=tab`).
- Test 5 asserted page-wide text, but the board grid behind the modal still renders every card — a false pass waiting to happen. Scoped to the palette dialog.

## Verification

- Both specs: **13 passed / 0 failed / 0 skipped**, three consecutive runs (was 4 passed / 9 skipped)
- ESLint clean on both files
- No product code changed — test-only

## Not masked

Two things I deliberately did **not** paper over:

1. The `[200, 204, 422]` tolerance was removed rather than kept — the test asserts the invariant holds when a second owner exists, so tolerating 422 would have contradicted its own name.
2. Where a route exists but the guard's premise was wrong, the guard is gone entirely rather than re-pointed at a status that happens to pass. `fetch` failures now throw.

## Remaining skips

4 dead-endpoint skips remain in `list-management` (method mismatches) and `presence-board` (routes may genuinely be GET-only), plus 3 UI-selector skips. `payment-card-price` keeps its 2 — covered by the standing payments exclusion. Those are the next batch.

### DIFF
--- a/tests/e2e/business-logic-invariants.spec.ts
+++ b/tests/e2e/business-logic-invariants.spec.ts
@@ -5,7 +5,7 @@
 // Based on: tests/e2e/business-logic-invariants-1.md (now deleted)
 
 import { test, expect } from '@playwright/test';
-import { BASE_URL, registerAndLogin, createWorkspace, createBoard, createList, createCard } from './_helpers';
+import { BASE_URL, registerAndLogin, registerAndGetCredentials, createWorkspace, createBoard, createList, createCard } from './_helpers';
 
 test.describe('Business Logic Invariants', () => {
   test.describe('Part 1 — Archived Board Read-Only Guard', () => {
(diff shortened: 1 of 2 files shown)


hermesrobinson-dev (approved):
Independent review APPROVE (head a0a74e3b). All five route claims verified against server source (not the diff): /api/v1/users/me is real and returns data.id; workspace and board search routes exist with the params used; unauthenticated workspace search returns 401; POST /workspaces/:id/members takes email, not userId. No masking: several assertions are strictly stronger than origin/main (the fail-loud fetch replaces a silent soft-skip; the [200,204,422] tolerance was tightened to 2xx + OWNER-remains; Test 3 still asserts both Gamma-absent and Alpha-present). ESLint exits 0 on both files.

PR #327, opened 2026-09-12
fix(e2e): remove dead-endpoint and UI-selector guards in list-management

matthewvryanjh:
## Summary

All four skipped tests in this spec had **never run**, and Test 8 asserted nothing meaningful. This fixes the routes, the selectors, and the two weak assertions.

## API causes

- **Reorder is `POST /boards/:id/lists/reorder`, not `PATCH`.** It also validates that the payload lists *every active list* on the board — it returned `reorder-count-mismatch` (400) otherwise. The old test sent 2 IDs against a board whose list count grew across the suite, so it would have failed the count check even once the method was right. It now uses a fresh board with a known count.
- **Archive is `PATCH /lists/:id/archive`, not `POST`.**

## UI causes

- Tests looked for `[data-testid="add-list"]`, `[data-testid^="list-header"]`, `[data-testid^="list-column"]` — **none exist**. The board renders a `+ Add a list` button; list headers are buttons where a leaf `Rename list <name>` button is nested inside a column button whose accessible name **concatenates its children**. Substring name matching hit both elements, so these use `exact: true`.
- `gotoBoard` now retries, because the app boot performs an async token refresh and navigating straight after login lands on `/workspaces`.

## Assertions strengthened, not weakened

- **Test 3** now confirms the reorder is actually persisted (fetches the board lists and checks the new index order) rather than only checking the status code.
- **New Test 3b** covers the count-mismatch rejection (400) — real product behaviour that had no coverage.
- **Test 8 (drag)** previously just checked "no crash" and skipped if it couldn't find columns. It now drags a real column and asserts the persisted order **changed**.
- **Tests 6/7** assert the new/renamed name appears in the UI *and* in the API, rather than trusting one surface.

## Helper fix

`createList` ignored its `title` argument and always auto-generated `List-<timestamp>`, so no caller could seed a list with a known name — which is why earlier attempts to locate a list by name failed. It now accepts an optional title (default behaviour unchanged).

## Verification

- **9 passed / 0 failed / 0 skipped**, three consecutive runs (was 5 passed / 4 skipped)
- ESLint clean on both files
- No product code changed — test-only

## Not masked

The three weak spots were tightened, not preserved: the `[200, 204]`-only status check now verifies persisted order, the drag test asserts a real order change instead of a smoke check, and the count-mismatch case is asserted rather than left untested.

### DIFF
--- a/tests/e2e/_helpers.ts
+++ b/tests/e2e/_helpers.ts
@@ -51,10 +51,11 @@ export async function createList(
   request: APIRequestContext,
   token: string,
   boardId: string,
+  title?: string,
 ): Promise<string> {
   const res = await request.post(`${BASE_URL}/api/v1/boards/${boardId}/lists`, {
     headers: { Authorization: `Bearer ${token}` },
-    data: { title: `List-${Date.now()}` },
+    data: { title: title ?? `List-${Date.now()}` },
   });
   const body = await res.json() as { data: { id: string } };
   return body.data.id;
--- a/tests/e2e/list-management.spec.ts
+++ b/tests/e2e/list-management.spec.ts
@@ -8,6 +8,22 @@ import { BASE_URL, registerAndGetCredentials, createWorkspace, createBoard, crea
 
 const UI_URL = process.env.TEST_UI_URL ?? 'http://localhost:5173';
 
+// The app boot performs an async token refresh, so navigating straight after
+// login can race it and land on /workspaces instead of the board.
+async function gotoBoard(page: import('@playwright/test').Page, boardId: string): Promise<void> {
+  for (let attempt = 0; attempt < 3; attempt++) {
+    await page.goto(`${UI_URL}/b/${boardId}`);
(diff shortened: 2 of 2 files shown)


hermesrobinson-dev (approved):
Independent review APPROVE (head aadae2a3). Routes and DOM verified against source: reorder is POST and validates order.length === activeLists.length (400 reorder-count-mismatch); archive is PATCH; create is 201; rename is PATCH returning data.id/data.title; GET board lists excludes archived. No masking — assertions are strictly stronger (persisted order checked, not just status; new count-mismatch rejection test; drag asserts a real order change; Tests 6/7 assert UI and API). The exact:true requirement is real: substring matching hits both the wrapping column button and the nested rename button (measured: substring 2, exact 1). Reviewer note on Test 8 (assertion detects first-vs-rest change, not arbitrary permutations) is sound but moot here — the test board has exactly two lists, so a first-list change is a complete permutation check.

PR #328, opened 2026-09-12
fix(e2e): remove presence guards for undelivered REST and UI surfaces

matthewvryanjh:
## Summary

Three tests in this spec had **never run**. Unlike PR #326/#327, none was a route-shape mistake — all three asserted **surface that does not exist**.

This PR therefore **removes** them rather than re-pointing them at passing statuses. That is a coverage decision, and it is the thing to scrutinise in review.

## Tests 2 and 3 — REST join/leave does not exist

They asserted `POST` and `DELETE /boards/:id/presence`. Presence is owned by the **realtime WebSocket layer**: a client subscribes to a board and the server sets `presence:<boardId>:<userId>`, expiring on disconnect.

Verified live:

| Request | Status |
|---|---|
| `GET /api/v1/boards/:id/presence` | 200 |
| `GET` unauthenticated | 401 |
| `POST /api/v1/boards/:id/presence` | **404** |
| `DELETE /api/v1/boards/:id/presence` | **404** |

Both paths 404 **by design**. The real join/leave behaviour is already covered by the remaining WS-driven test, which subscribes two clients and reads the list back over REST. A test asserting a 404 would be testing nothing, so these are deleted rather than made to pass.

## The UI test — an undelivered feature

It asserted a second user's avatar appears in the first user's board header.

`src/extensions/Realtime/components/PresenceAvatars.tsx` **exists and is complete but is never imported.** `grep -rn "from.*PresenceAvatars" src/` returns only a type-only import of `PresenceUser` — never the component. The board header renders the static `MemberAvatarStack` instead.

So the test could never pass. It soft-skipped on every run under a selector guard, which is why the gap was invisible in the suite output.

**Wiring `PresenceAvatars` into the board header is a product change, not a test fix.** It is out of scope here and recorded in the E2E recovery backlog as a product item.

## Why removing rather than leaving it failing

Keeping a permanently-failing test would break CI for an unrelated future change. Keeping it skipping is what hid the gap for this long. The gap is now recorded where it will be worked, and the test can be written properly against the real component when it ships.

## Verification

- **3 passed / 0 failed / 0 skipped**, three consecutive runs (was 3 passed / 3 skipped)
- ESLint clean
- No product code changed

## Reviewer focus

1. Confirm `PresenceAvatars` is genuinely never imported (don't take the PR body's word).
2. Confirm POST/DELETE really are absent routes in `server/extensions/presence/api/index.ts`, and that the WS path is already covered by the remaining test — i.e. that no real coverage is lost.
3. Say so if you think the UI test should be kept as a deliberately failing marker instead of removed.

### DIFF
--- a/tests/e2e/presence-board.spec.ts
+++ b/tests/e2e/presence-board.spec.ts
@@ -168,22 +168,35 @@ test.describe('Board Presence', () => {
       return body.data.map((u) => u.id);
     };
 
+    // Resolve both user ids so the assertion can name WHO left rather than only
+    // that the list shrank.
+    const idOf = async (token: string): Promise<string> => {
+      const res = await request.get(`${BASE_URL}/api/v1/users/me`, {
+        headers: { Authorization: `Bearer ${token}` },
+      });
+      expect(res.status()).toBe(200);
+      const body = await res.json() as { data: { id: string } };
+      return body.data.id;
+    };
+    const idA = await idOf(tokenA);
+    const idB = await idOf(tokenB);
+
     const socketA = await openSocket(tokenA);
     const socketB = await openSocket(tokenB);
 
     try {
-      // Both subscribers are recorded.
+      // Both subscribers are recorded, by id.
       await expect
         .poll(readPresenceIds, { timeout: 8000 })
-        .toEqual(expect.arrayContaining([expect.any(String)]));
-      const both = await readPresenceIds();
-      expect(both.length).toBeGreaterThanOrEqual(2);
+        .toEqual(expect.arrayContaining([idA, idB]));
 
-      // B leaves; the presence list must shrink to exclude B.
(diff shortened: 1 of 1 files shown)


hermesrobinson-dev (approved):
Independent reviews APPROVE (head 521ae962). Commit 1: verified POST/DELETE /boards/:id/presence genuinely 404 by design (presence is WebSocket-driven), and that the surviving WS test covers join; confirmed PresenceAvatars.tsx is never imported (grep returns only a type-only PresenceUser import) while BoardHeader renders MemberAvatarStack — so the deleted UI test asserted an undelivered feature. Commit 2 added leave coverage; commit 3 tightened it to assert identity (B gone, A remains) rather than only a smaller count, and was proven non-vacuous by fault injection (disabling cache.del in unsubscribe.ts makes it fail). No masking; deletion of the three never-running tests loses no coverage.

PR #329, opened 2026-09-12
test(e2e): presence join asserts viewer identity; correct over-broad comment

matthewvryanjh:
## Summary

Two accuracy defects flagged by the #328 review, both now live on `main`. Neither was a merge blocker, but both were real:

## 1. An over-broad comment

The comment above the removed REST tests claimed:

> the real join/leave behaviour is covered by Test 2 below

Test 2 covers **join only** — leave is covered by Test 3. The comment now names both tests correctly.

## 2. The join test did not assert identity

It asserted `body.data.length >= 2`, which proves only that *two viewers exist* — not that the two specific subscribers are the ones listed. It now resolves both user ids via `/api/v1/users/me` and asserts the presence list contains **both by id**.

## 3. Duplication removed

Test 2 and Test 3 carried byte-identical inline WebSocket subscribe helpers. One is hoisted to module scope.

## Proof the assertions are not vacuous

This is the part that matters — an assertion that can never fail is the exact problem this whole batch of work exists to fix. So I fault-injected against the real server:

| Injected fault | Result |
|---|---|
| Disable the presence **write** in `rooms/subscribe.ts` | join test fails: `Expected: ArrayContaining [<idA>, <idB>]` / `Received: []` |
| Disable the presence **delete** in `rooms/unsubscribe.ts` | leave test fails: `Expected: false` / `Received: true` |

Both product files were restored afterwards (`git diff server/` is empty), and the suite re-run clean to confirm.

## Verification

- **4 passed / 0 failed / 0 skipped**, three consecutive runs
- ESLint clean
- No product code changed

### DIFF
--- a/tests/e2e/presence-board.spec.ts
+++ b/tests/e2e/presence-board.spec.ts
@@ -9,6 +9,38 @@ import { BASE_URL, registerAndGetCredentials, createWorkspace, createBoard, crea
 
 const WS_URL = (process.env.TEST_BASE_URL ?? 'http://localhost:3000').replace(/^http/, 'ws');
 
+// Open a realtime WebSocket, subscribe to the board, and resolve once the server
+// accepts the subscription. Used by the join and leave tests.
+function subscribeToBoard(token: string, boardId: string): Promise<WebSocket> {
+  return new Promise((resolve, reject) => {
+    const ws = new WebSocket(`${WS_URL}/api/v1/ws?token=${encodeURIComponent(token)}`);
+    const timer = setTimeout(() => reject(new Error('WS subscribe timed out')), 8000);
+    ws.addEventListener('open', () => {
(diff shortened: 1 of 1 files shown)


hermesrobinson-dev (approved):
Independent review APPROVE (head 561389f3). Verified: the corrected comment now attributes join to Test 2 and leave to Test 3, matching the code; the join assertion is genuinely stronger (arrayContaining requires BOTH ids, so it cannot pass with only one subscriber present, unlike the previous length>=2); the hoisted subscribeToBoard helper is equivalent to the two inline copies with no behaviour dropped (still rejects on WS error and on timeout); test count unchanged at 4 runnable. The fault-injection claim is plausible and consistent: handleGetPresence reads the presence:<boardId>:<userId> cache keys, so skipping cache.set in subscribe.ts would yield Received: []. ESLint exits 0.

PR #330, opened 2026-09-12
test(e2e): close presence sockets on failure and drain teardown

matthewvryanjh:
## Summary

Two socket-hygiene nits from the #328 review. Both are flake sources rather than correctness bugs, but this spec creates real WebSocket connections and runs against shared state, so they matter.

## 1. Sockets leaked on every rejection path

`subscribeToBoard` rejected on timeout, on a WS `error` frame, and on the `error` event — **without closing the WebSocket it created**. Each leak holds an open connection holding a `presence:<boardId>:<userId>` key.

Worse: if the *second* subscribe failed, the caller never reached its `finally`, so the *first* socket leaked too. That is exactly the shape of a test that leaves a phantom viewer behind.

Every rejection path now closes the socket.

## 2. Teardown raced the next test

Tests 2 and 3 share `boardId` and both tokens from `beforeAll`. Their `finally` blocks called `socket.close()` — which only *initiates* the close — without awaiting the server-side async unsubscribe. So a late `cache.del` could land while the next test was subscribing on the same board, dropping its presence count and failing nondeterministically.

Added `closeAndDrain()`: closes a socket and polls `GET /boards/:id/presence` until that viewer's id is gone. Both tests drain in `finally`, so a test cannot leave state behind.

This is the false-*failure* direction — a flake, not a masked pass — but it would have shown up as an intermittent red run, which is how test suites lose trust.

## Verification

- **4 passed / 0 failed / 0 skipped, five consecutive runs** (most runs I've done on this spec)
- ESLint clean
- No product code changed

### DIFF
--- a/tests/e2e/presence-board.spec.ts
+++ b/tests/e2e/presence-board.spec.ts
@@ -4,7 +4,7 @@
 // Based on: specs/tests/presence-board.md
 // Soft-skips when the server is not reachable.
 
-import { test, expect } from '@playwright/test';
+import { test, expect, type APIRequestContext } from '@playwright/test';
 import { BASE_URL, registerAndGetCredentials, createWorkspace, createBoard, createList, type Credentials } from './_helpers';
 
 const WS_URL = (process.env.TEST_BASE_URL ?? 'http://localhost:3000').replace(/^http/, 'ws');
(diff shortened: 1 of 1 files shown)


hermesrobinson-dev (approved):
Independent reviews APPROVE both commits. Commit 1 (1fd561e0): sockets now close on every rejection path; TDZ risk confirmed not real (WebSocket events are always queued as tasks, so timer is assigned before fail can run). Commit 2 (ea7aaf21): sync-throw handler correct and leaves no dangling socket/timer/listener; settled guard is idempotent and no hang path exists (every write of settled is followed by resolve or reject). closeAndDrain correct; no assertion weakened.

PR #331, opened 2026-09-12
test(e2e): drag reorder asserts the change persisted to the API

matthewvryanjh:
## Summary

The #327 review flagged that the drag test asserted **UI order only**. A drag that mutated local React state but never called the reorder endpoint would have passed — so a persistence regression could slip through. That is the same class of gap as the soft-skips this batch exists to remove.

## The two layers

In `BoardCanvas.tsx`, `onListReorder` updates local state; `onDragCommit` is what hits the reorder endpoint and persists. The old assertion read only the DOM, so it could not distinguish them.

## What changed

Test 8 now:

1. Captures the API list order **before** the drag and asserts the seeded order as a precondition (so a wrong starting order fails loudly rather than silently).
2. Asserts the UI order changed.
3. **Polls `GET /boards/:id/lists` until it shows the swapped order** — the persistence check that was missing.

## Proven non-vacuous by fault injection

Replacing `await onDragCommit(...)` in `BoardCanvas.tsx` with a no-op — local state only, nothing persisted — makes the test **fail** on a deep-equality mismatch of the API order.

The product file was restored afterwards; `git diff src/ server/` is empty.

This is the third assertion in this batch verified this way. "The test goes green" is not evidence of coverage — it was exactly the evidence that hid 16 dead tests — so a new assertion gets fault-injected before I claim it works.

## Verification

- **9 passed / 0 failed / 0 skipped**, three consecutive runs
- ESLint clean
- No product code changed

### DIFF
--- a/tests/e2e/list-management.spec.ts
+++ b/tests/e2e/list-management.spec.ts
@@ -244,8 +244,22 @@ test.describe('List Management', () => {
     const uiToken = uiCreds.token;
     const uiWorkspaceId = await createWorkspace(request, uiToken);
     const uiBoardId = await createBoard(request, uiToken, uiWorkspaceId);
-    await createList(request, uiToken, uiBoardId, `Drag A ${run}`);
-    await createList(request, uiToken, uiBoardId, `Drag B ${run}`);
+    const nameA = `Drag A ${run}`;
+    const nameB = `Drag B ${run}`;
+    await createList(request, uiToken, uiBoardId, nameA);
+    await createList(request, uiToken, uiBoardId, nameB);
+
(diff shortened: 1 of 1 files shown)


hermesrobinson-dev (approved):
Independent review APPROVE (head 33e4e762). Critical check passed: GET /api/v1/boards/:id/lists does orderBy('position','asc') (list.ts:39-41), so the assertion reads genuinely persisted order, not insertion order; and reorder.ts writes the same position column. Scope confirmed limited to Test 8 with no other assertion weakened. Non-vacuous per the fault-injection check (no-oping onDragCommit makes it fail), consistent with onListReorder updating local state and onDragCommit being the persistence path.

